PR #89, opened 2023-12-12
chore: migrate to farmOS v3.x and postgres v13

### LINKED ISSUE #64: farmOS 3.x Migration
This is a place to collect issues that need attention related to the migration to farmOS 3.x.

Much of the following is from the [release notes](https://github.com/farmOS/farmOS/blob/3.0.0-beta2/CHANGELOG.md).

- [x] -  "farmOS v3 updates Drupal to version 10. Drupal 9 is end-of-life as of November 1st, 2023. If you have built any add-on modules for farmOS, you will need to check that they are compatible with Drupal 10, and declare support in your *.info.yml file by changing core_version_requirement from ^9 to ^9 || ^10 (to indicate that it works on both versions), or just ^10 (to indicate that it only works on Drupal 10).
  - Should be a fairly easy update and should go along with requiring the "farmOS Inventory" module and the "Simple OAuth Password Grant" module.
- [x]  - If you are using PostgreSQL, Drupal 10 requires PostgreSQL version 12 or greater, with the pg_trgm extension enabled. If you have PostgreSQL 13 or greater, the pg_trgm extension will be enabled automatically during farmOS installation. PostgreSQL 12 users, or users who are updating from farmOS 2.x to 3.x, will need to enable it manually by running the following query after the farmOS database has been created, but before farmOS is installed/updated: CREATE EXTENSION pg_trgm;
  - The postgress image in the `FD2-Images` repo should just be updated to use PostgresSQL 13.  This may or may not require rebuilding the base and sample databases in the `FD2-SampleDB` repo.
- [x] - The biggest changes are that the "Implicit" grant type has been removed
  - **This is the big unknown... FarmData2 needs to be able to tell if a user has permissions to do particular operations (e.g. create logs vs create terms vs just viewing things).  How is that handled/determined in light of this change?**
- [x] - the "Password Credentials" grant type has been moved to an optional "Simple OAuth Password Grant" module, which must be enabled in order to use that grant type.
  - Will want this to be enabled in the development environment.  So add something to the `buildSampleDB.bash` script in the `FD2-SampleDB` repo.
- [x] - The default farmOS client that is included with farmOS has also been moved to a separate module that is not enabled by default. 
  - Probably want to just go ahead and create a custom consumer for the FarmData2 development environment to use.  This would replace the `addConsumerCORSorigins.cy.js` script  in the `FD2-SampleDB` repo. that is currently used in creating the sample database.  It would create a full consumer instead of just augmenting the default `farm` consumer. Then the code that uses the `farm` consumer should change to use the custom FarmData2 consumer.
- [x] - Will need to update the farmOS docker image produced by the `FD2-Images` repo.

braughtg:
**Pull Request Description**

This PR updates the `docker/docker-compose.yml` file to use new images for the farmOS and postgres.
- The farmOS image updates to v3.x.
- The postgres image updates to v13 (required by farmOS v3.x)

Closes #64

---

**Licensing Certification**


braughtg:
This PR was created from the incorrect branch.

PR #114, opened 2023-12-20
fix: integrates missing release commit from production

braughtg:
**Pull Request Description**

When creating the first (1.0.0) release by pushing production, the commit created by the semantic release bot was never pulled into the development branch. The development branch was rebased onto production locally and this is a PR of that rebased development branch.  The idea is just to see if this will allow development to be updated to where it can be `--ff-only` merged into production without corrupting the development branch and forcing other to `--force` pull.

---

**Licensing Certification**


braughtg:
This seems like it will work, but will tag every change with this PR and thus a lot of context will be lost.  At this point there are very few forks out there.  So a force push to development will require them to force pull it.  But that seems to be the better option at this point in time.

PR #144, opened 2024-01-23
Added Bash Trap and Cleanup Function to test.bash

### LINKED ISSUE #141: Ctrl-C from `test.bash` leaves zombie processes
The `test.bash` script starts up the necessary servers when it is run.  They are started in a process group and the script terminates them when it completes.  However, if the script is terminated using `Ctrl-C` then the part that terminates the servers does not run.  The effect is that there can be a bunch of zombie servers running. This creates stability issues with Cypress and also bogs down the dev environment.

I'm not sure what the solution is here... maybe its possible to tie this process group to the process running `test.bash` so that when it exits they are also terminated?

WarpWing:
**Pull Request Description**

Fixes `Ctrl+c` failures to close out "zombie" servers after failed tests. 

Closes #141 

**Licensing Certification**

### DIFF
--- a/bin/test.bash
+++ b/bin/test.bash
@@ -73,7 +73,7 @@ cleanup() {
     echo "Builder terminated."
   fi
   # Exit the script
-  exit 130
+  exit $EXIT_CODE
 }
 # Trap SIGINT (Ctrl-C) and SIGTERM
 trap cleanup SIGINT SIGTERM
@@ -340,4 +340,4 @@ fi
 echo "Tests complete."
 
 # Runs normal cleanup
-cleanup
+cleanup "$EXIT_CODE"
--- a/bin/test.bash
+++ b/bin/test.bash
@@ -41,6 +41,43 @@ function usage {
   exit 255
 }
 
+# Define cleanup() function. This needs to be defined ASAP to catch SIGINT and SIGTERM at the earliest signs of trouble.
+cleanup() {
+  echo "Cleaning up..."
+
+  # Terminate the dev server if it's running
+  if [ -n "$DEV_GID" ]; then
+    echo "Terminating the dev server..."
+    kill -INT -- -"$DEV_GID"
+    echo "Dev server terminated."
+  fi
+
+  # Terminate the builder if it's running
+  if [ -n "$BUILDER_GID" ]; then
+    echo "Terminating the builder..."
+    kill -INT -- -"$BUILDER_GID"
+    echo "Builder terminated."
+  fi
+
+  # Terminate the preview server if it's running
+  if [ -n "$PREVIEW_GID" ]; then
+    echo "Terminating preview server..."
+    kill -INT -- -"$PREVIEW_GID"
+    echo "Preview server terminated."
+  fi
+
+  # Terminate the live builder if it's running
+  if [ -n "$LIVE_GID" ]; then
+    echo "Terminating the builder..."
+    kill -INT -- -"$LIVE_GID"
+    echo "Builder terminated."
+  fi
+  # Exit the script
+  exit "$EXIT_CODE"
+}
+# Trap SIGINT (Ctrl-C) and SIGTERM
+trap cleanup SIGINT SIGTERM
+
 # Change into the main repo directory.
 SCRIPT_PATH=$(readlink -f "$0")                     # Path to this script.
 SCRIPT_DIR=$(dirname "$SCRIPT_PATH")                # Path to directory containing this script.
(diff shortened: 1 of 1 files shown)


WarpWing:
I would like to note some of my concerns that I'm not sure should be concerns but zombie processes might still remain if any of the child processes of the servers fork their own children that do not terminate on SIGINT. Also that killing process groups with `kill -INT -- -"$GID"` sends a SIGINT to all processes in the group, which may not terminate all of them if they don't handle SIGINT properly. That is just speculation due to me not being too familiar with the FarmData2 Codebase. I'd imagine it wouldn't be a problem if it is, I can look into remedying that.

bin/test.bash
```diff
@@ -41,6 +41,43 @@ function usage {
...
+  fi
+
+  # Terminate the live builder if it's running
+  if [ -n "$LIVE_GID" ]; then
+    echo "Terminating the builder..."
+    kill -INT -- -"$LIVE_GID"
+    echo "Builder terminated."
+  fi
+  # Exit the script
+  exit "$EXIT_CODE"
```
braughtg: The $EXIT_CODE is not set here when `cleanup` is called by `trap`.  So I'm seeing this error when I use `Ctrl-C` to terminate `test.bash`:

image

Maybe pass the `$EXIT_CODE` in as a parameter from line 343 and use a default of `130` (Terminated by `Ctrl-C`) if `cleanup` is called by the `trap`?

braughtg (changes requested):
This is close.  It properly traps the `Ctrl-C` and terminates the process group. But it also generates an error when `Ctrl-C` is used.  See my comment in the code on line 76.

braughtg:
> I would like to note some of my concerns that I'm not sure should be concerns but zombie processes might still remain if any of the child processes of the servers fork their own children that do not terminate on SIGINT. Also that killing process groups with `kill -INT -- -"$GID"` sends a SIGINT to all processes in the group, which may not terminate all of them if they don't handle SIGINT properly. That is just speculation due to me not being too familiar with the FarmData2 Codebase. I'd imagine it wouldn't be a problem if it is, I can look into remedying that.

Good points!  It appears that, at least right now, all of the processes in the process group respond appropriately to `SIGINT`.  So I'm comfortable going forward with this approach at this point.

bin/test.bash
WarpWing: > The $EXIT_CODE is not set here when `cleanup` is called by `trap`. So I'm seeing this error when I use `Ctrl-C` to terminate `test.bash`:
> 
> image
> 
> Maybe pass the `$EXIT_CODE` in as a parameter from line 343 and use a default of `130` (Terminated by `Ctrl-C`) if `cleanup` is called by the `trap`?

I was worried about that. I was considering using the standard 0 but 130 should be fine. I wasn't sure what exit code to use, I should've had the foresight to realize that. Let me change it 130 real quick.

WarpWing: Exit Code changed.

WarpWing (commented):
Exit code has been changed to 130

WarpWing:
On second thought, I just re-read and I can modify the function to pass in the exitcode as a parameter.

bin/test.bash
```diff
@@ -41,6 +41,43 @@ function usage {
...
+  fi
+
+  # Terminate the live builder if it's running
+  if [ -n "$LIVE_GID" ]; then
+    echo "Terminating the builder..."
+    kill -INT -- -"$LIVE_GID"
+    echo "Builder terminated."
+  fi
+  # Exit the script
+  exit 130
```
braughtg: This should not exit with `130` all of the time.  If Cypress is terminated normally and the script enters cleanup from line 343 then the exit status should be as it was set on 333 or 337.  So maybe pass it as a parameter to `cleanup` from 343?

braughtg (changes requested):

bin/test.bash
WarpWing: That would make sense, I'll modify cleanup with a parameter and define it as 130 as default but will be overriden by any passed through exit code.

bin/test.bash
```diff
@@ -73,7 +73,7 @@ cleanup() {
     echo "Builder terminated."
   fi
   # Exit the script
-  exit 130
+  exit $EXIT_CODE
```
braughtg, line 79: There is still an issue here... the `$EXIT_CODE` is still going to be undefined if `cleanup` is called from a `Ctrl-C`.  You aren't seeing the same error as earlier because `$EXIT_CODE` is not quoted (`"$EXIT_CODE"`).  I don't so much care about the quotes here as _string splitting_ shouldn't be an issue.  But the exit code should be `130` in this case.  So maybe just add a conditional here that if `$EXIT_CODE` is not defined then set it to `130`.

WarpWing: > There is still an issue here... the `$EXIT_CODE` is still going to be undefined if `cleanup` is called from a `Ctrl-C`. You aren't seeing the same error as earlier because `$EXIT_CODE` is not quoted (`"$EXIT_CODE"`). I don't so much care about the quotes here as _string splitting_ shouldn't be an issue. But the exit code should be `130` in this case. So maybe just add a conditional here that if `$EXIT_CODE` is not defined then set it to `130`.

Yeah, I'll get to adding a conditional before putting it up for review, I don't know why I added the "" to the exit code, I honestly think I was mentally autopiloting but I am curious if there's a way to pass in default parameters like in Python (though Bash is for scripting, less for programming)

WarpWing: Added Ad-Hoc conditional.

WarpWing:
I added a little parameter expansion that sets `$EXIT_CODE` to 130 if no value is initialized.

braughtg (approved):
Looks good!  Thanks for you work on this!

PR #269, opened 2024-05-23
improves direct_seeding.submission.e2e.cy.js tests

### LINKED ISSUE #170: Form Submission e2e Test Improvement
Currently the unit tests for the `submit` function in `lib.js` for each entry point (Tray Seeding / Direct Seeding) thoroughly test that the correct assets, quantities and logs get created in the farmOS database based on the `form` that they are passed.

In normal use, the `form` is created by the UI and passed to the `submit` function.  Currently, the e2e tests for each entry point test that the `submit` function is called by intercepting the final API request that it makes.  They do not however test that all of the assets, quantities and logs are created.  They should not have to, because `submit` is fully tested by the unit tests.  

However, these e2e tests are currently unable to efficiently ensure that all of the correct information is in the form that is passed to `submit`.  The best way to do this would be to _spy_ on the `submit` function and simply check the contents of the argument that is passed. However, I have not been able to get a spy to work on this function due to import issues.  More specifically, the test would need to import `lib.js` which it can do.  However, `lib.js` imports `@libs/farmosUtil/farmosUtil` where `@libs` is defined in the `vite.config.js` file, and for some reason that import is failing to resolve.

It would be good to have some efficient way to test that `submit` is getting all of the information that is is supposed to get.  Take _efficient_ here to mean not asking the database for all of the assets, quantities and logs.

It seems like there should be a good way to do this with `cy.spy` but the build issues are currently a show stopper.

When a solution has been found it should be back ported to:
- [x] - Tests for tray Seeding `lib.js`
- [x] - Tests for Direct Seeding `lib.js`
- [x] - Tests for Transplanting `lib.js` - if it has been written.

sidlamsal:
**Pull Request Description**

Adds a test using a spy on the submitForm function to check that it was called with the correct data.

Partially addresses #170 

Co-authored-by: Shahir Ahmed <98346408+Shahir-47@users.noreply.github.com>

---

**Licensing Certification**

### DIFF
--- a/modules/farm_fd2/src/entrypoints/direct_seeding/direct_seeding.submission.e2e.cy.js
+++ b/modules/farm_fd2/src/entrypoints/direct_seeding/direct_seeding.submission.e2e.cy.js
@@ -110,8 +110,6 @@ describe('Direct Seeding: Submission tests', () => {
       expect(spy).to.be.calledOnce;
 
       let formData = spy.getCall(0).args[0];
-      console.log(formData);
-      console.log('formData');
       expect(formData.seedingDate).to.equal('1950-01-02');
       expect(formData.cropName).to.equal('ARUGULA');
       expect(formData.locationName).to.equal('ALF');
--- a/modules/farm_fd2/src/entrypoints/direct_seeding/direct_seeding.submission.e2e.cy.js
+++ b/modules/farm_fd2/src/entrypoints/direct_seeding/direct_seeding.submission.e2e.cy.js
@@ -1,4 +1,41 @@
+import * as farmosUtil from '../../../../../library/farmosUtil/farmosUtil.js';
+
 describe('Direct Seeding: Submission tests', () => {
+  let cropMap = null;
+  let fieldMap = null;
+  let equipMap = null;
+  let bedMap = null;
+  let valuesMap = null;
+  let maps = new Map();
+
+  before(() => {
+    // get maps
+    let cropPromoise = farmosUtil.getCropNameToTermMap();
+    let fieldPromise = farmosUtil.getFieldNameToAssetMap();
+    let equipmentPromise = farmosUtil.getEquipmentNameToAssetMap();
+    let bedPromise = farmosUtil.getBedNameToAssetMap();
+
+    valuesMap = new Map();
+    valuesMap.set('Bed Feet', '200');
+    valuesMap.set('Rows/Bed', '5');
+    valuesMap.set('Row Feet', '1000');
+    valuesMap.set('Bed Width', '30');
+    valuesMap.set('Depth', '6');
+    valuesMap.set('Speed', '5');
+
+    let promiseMap = new Map();
+    promiseMap.set('crop', cropPromoise);
+    promiseMap.set('field', fieldPromise);
+    promiseMap.set('equipment', equipmentPromise);
+    promiseMap.set('bed', bedPromise);
+
+    return promiseMap.forEach((values, keys) =>
+      values.then((map) => {
+        maps.set(keys, map);
+      })
+    );
+  });
+
   beforeEach(() => {
     cy.restoreLocalStorage();
     cy.restoreSessionStorage();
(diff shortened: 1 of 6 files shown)


modules/farm_fd2/src/entrypoints/direct_seeding/direct_seeding.submission.e2e.cy.js
```diff
@@ -1,4 +1,41 @@
...
+
 describe('Direct Seeding: Submission tests', () => {
+  let cropMap = null;
+  let fieldMap = null;
+  let equipMap = null;
+  let bedMap = null;
+  let valuesMap = null;
+  let maps = new Map();
+
+  before(() => {
```
braughtg: The `before` should not `return` a value.  You have it returning the `Map` of `Promises` from the fetches of the maps that you use in the test.  Instead, I think you should have some code to fetch each map that looks similar to:

```JavaScript
farmosUtil.getFieldNameToAssetMap().then((map) => {
      fieldMap = map;
    }
```

Also, the `promiseMap` doesn't seem to really be used for anything other than assigning the values and so it seems overly complicated for that purpose as compared to the above.  If the `promiseMap` is eliminated then the test can just use the map variables directly.

modules/farm_fd2/src/entrypoints/direct_seeding/direct_seeding.submission.e2e.cy.js
```diff
@@ -81,7 +118,43 @@ describe('Direct Seeding: Submission tests', () => {
      * lib.submitForm function that is called when the "Submit"
      * button is clicked.
      */
-    cy.intercept('POST', '**/api/log/activity', cy.spy().as('submitSpy'));
+    cy.intercept('POST', '**/api/asset/plant', (req) => {
+      // check plant name
+      cropMap = maps.get('crop');
+      let cropID = cropMap.get('ARUGULA').id;
+      expect(req.body.data.relationships.plant_type.data[0].id).to.eq(cropID);
+    }).as('submitSpy');
```
braughtg: In order to use the `submitSpy` later to check the number of times the intercept happened (e.g. `cy.get('@submitSpy').should('have.callCount', 3)`) then you will need to give each one of them a different alias (e.g. maybe call this one `plantAssetSpy`)

modules/farm_fd2/src/entrypoints/direct_seeding/direct_seeding.submission.e2e.cy.js
```diff
@@ -81,7 +118,43 @@ describe('Direct Seeding: Submission tests', () => {
      * lib.submitForm function that is called when the "Submit"
```
braughtg: Be sure to edit/delete/move/clean up comments so that they agree with the way your code works now.

braughtg (dismissed):
This did all run and pass for me.  Which is a good start!  

I added some comments that I think will both improve the readability / understandability of the code.  As you read those and think about revision, try to imagine coming back to this in a year when a test fails and trying to understand what happened?  Then try to revise so that the test code will be as easy to understand as you can make it.

Possibly another way to make the code more understandable is to think about the goal of this test being to check that each piece of `form.data` is being passed as expected.  From this perspective we only need to check that each piece of `form.data` appears correctly in one `req.body`.  This is because the `lib.submit.unit.cy.js` test ensures that if `form.data` is correct then the assets, logs, quantities created are correct.  So maybe thinking about and organizing this test using that principle will help with readability and cohesion?

braughtg:
Also, be sure to add a `co-author` footer to the PR so that both @sidlamsal and @Shahir-47 are credited for the commit when it is merged.  You can find information about the `co-author` footer in part 8 of the `CONTRIBUTING.md` documentation.

braughtg:
@sidlamsal @Shahir-47 Change of plans... I was able to get a `spy` working on the `lib.submitForm` function that will simplify the `*.submission.e2e.cy.js` tests.  So you should now revise your test to follow this model.  Sorry for the time spent on the other approach.  I think what was learned about the larger farmOS / FarmData ecosystem will still be valuable.

You can find the details in this comment: https://github.com/FarmData2/FarmData2/issues/170#issuecomment-2129875076

I have also implemented this in the Tray Seeding entry point and merged it to development with PR #270. So you can see it in context there if that helps.

modules/farm_fd2/src/entrypoints/direct_seeding/direct_seeding.submission.e2e.cy.js
```diff
@@ -173,11 +103,25 @@ describe('Direct Seeding: Submission tests', () => {
...
-     * already do that.
+     * Check that the submitForm function was called with the
+     * expected form data.
      */
-    //cy.get('@submitSpy').should('be.calledOnce');
+    cy.get('@submitFormSpy').then((spy) => {
+      expect(spy).to.be.calledOnce;
+
+      let formData = spy.getCall(0).args[0];
+      console.log(formData);
```
braughtg: These `console.log` statements should be removed.

braughtg (changes requested):
Overall this looks good.  Address the one comment, then review your work one last time as a pair.  When you are happy with it, mark it as ready for review.  I will then pull, test and hopefully merge it!

braughtg (approved):
Nicely done!

braughtg:
Can this be reopened?

Issue #223, opened 2024-04-09
Transplanting e2e tests


braughtg:
The transplanting entry point needs to have e2e tests written.

Use the guidance in `modules/README.md` and the e2e tests in the tray seeding and direct seeding modules as a guide.  The tests for submission of the form should be sure to use the `spy` approach as detailed in #170 and #270.

braughtg:
Closed by #271

Issue #222, opened 2024-04-09
Transplanting: Needs to have unit tests for `submitForm` in `lib.js`


braughtg:
The `submitForm` function in `lib.js` in the transplanting entry point needs to be tested.

Use the tests for `lib.js` in the tray seeding and direct seeding modules as a model.

braughtg:
Closed by #271

PR #271, opened 2024-05-28
Removes area from transplant form

### LINKED ISSUE #265: Modify the Transplanting form to not include area
This issue will require:
  - [ ] - Change the form in `App.vue` to use the new prop to hide the area field.
  - [ ] - Update `submitForm` in `lib.js` to not create a quantity for area.
  - [ ] - Write the e2e tests for the Transplanting form (see #223).
  - [ ] - Write the `lib.js` tests to not check for the area quantity (see #222)

Shahir-47:
**Pull Request Description**

Modifies the Transplanting form to not include area

Closes #265 #223 #222 #170

Co-authored-by: sidlamsal <98344853+sidlamsal@users.noreply.github.com>

---

**Licensing Certification**


Shahir-47:
Even though overlay is visible when the icon is clicked, data-cy="picklist-info-user-0" does not exist. Here is the code snippet I tried:
```
    cy.get('[data-cy="picklist-info-icon-0"]').should('exist').click();
    cy.get('[data-cy="picklist-info-overlay"]').should('be.visible');
    cy.get('[data-cy="picklist-info-user-0"]').should('exist');
    cy.get('[data-cy="picklist-info-user-0"]').should(
      'have.text',
      'User: admin'
    );
```

modules/farm_fd2/src/entrypoints/transplanting/transplanting.submitReset.e2e.cy.js
```diff
@@ -0,0 +1,227 @@
...
+    cy.get('[data-cy="comment-input"]').type('Test comment');
+    cy.get('[data-cy="comment-input"]').blur();
+  }
+
+  it('Invalid date disables submit', () => {
+    populateForm();
+    cy.get('[data-cy="submit-button"]').should('be.enabled');
+    cy.get('[data-cy="date-input"]').clear();
+    cy.get('[data-cy="submit-button"]').click();
+    cy.get('[data-cy="submit-button"]').should('be.disabled');
```
braughtg, line 80: There is improvement to these tests that I have thought of since writing the originals in Tray Seeding and Direct seeding.  After checking that an invalid value disables the submit button, the value should be changed back to a valid value to ensure that the submit button is then re-enabled.  This addition ensures that the logic that re-enables the button works as well as the logic that disables it.

This small addition should be added to each of the tests in this file.

braughtg (changes requested):
Overall this looks good. I suggested an improvement to the `SubmitReset` tests that I thought of since writing the original tests.

braughtg:
Looks like this PR closes a few issues.  Be sure to list them all in the main comment at the top.

modules/farm_fd2/src/entrypoints/transplanting/transplanting.crop.e2e.cy.js
```diff
@@ -0,0 +1,65 @@
...
+  it('Crop props are correct', () => {
+    cy.get('[data-cy="transplanting-picklist"]')
+      .find('[data-cy="selector-required"]')
+      .should('be.visible');
+    cy.get('[data-cy="transplanting-picklist"]')
+      .find('[data-cy="selector-input"]')
+      .should('have.value', null);
+  });
+
+  it('Crop validity styling works', () => {
```
braughtg: There are a couple of cases you should handle here.  In one there is no crop selected which is what you have tested. In another there is a crop selected but no trays have been selected for transplanting. You should add another test to get this second case.

braughtg:
> Even though overlay is visible when the icon is clicked, data-cy="picklist-info-user-0" does not exist. Here is the code snippet I tried:
> 
> ```
>     cy.get('[data-cy="picklist-info-icon-0"]').should('exist').click();
>     cy.get('[data-cy="picklist-info-overlay"]').should('be.visible');
>     cy.get('[data-cy="picklist-info-user-0"]').should('exist');
>     cy.get('[data-cy="picklist-info-user-0"]').should(
>       'have.text',
>       'User: admin'
>     );
> ```

This test passes in the Electron browser but not in the Firefox browser.  The issue seems to be with how the two different browsers render their scroll bars.  In Firefox the `info-icon` is partially obscured by the scroll bar so it cannot be clicked. While in Electron the scroll bar does not obscure the `info-icon`.  The solution here is to use a "force click" which will allow an element to be clicked in a test without it being directly visible and is useful in cases like this.  

Here's an example:
```Javascript
    cy.get('[data-cy="picklist-info-icon-0"]')
      .should('exist')
      .click({ force: true });
```

modules/farm_fd2/src/entrypoints/transplanting/transplanting.submitReset.e2e.cy.js
```diff
@@ -78,20 +78,47 @@ describe('Transplanting: Submit/Reset Buttons component', () => {
...
+      .find('[data-cy="picker-options"]')
+      .find('input')
+      .eq(0)
+      .click();
+    cy.get('[data-cy="transplanting-location"]')
+      .find('[data-cy="picker-options"]')
+      .find('input')
+      .eq(3)
+      .click();
+    cy.get('[data-cy="transplanting-location"]').find(
```
braughtg: This statement doesn't seem to do anything but find an element.  Should it do something with the element? Or is it unnecessary?

braughtg (changes requested):
Looking good.  Just one small comment to have a look at.

modules/farm_fd2/src/entrypoints/transplanting/App.vue
```diff
@@ -182,7 +182,7 @@
               v-bind:equipment="form.equipment"
               v-bind:depth="form.depth"
               v-bind:speed="form.speed"
-              v-bind:area="form.area"
+              v-bind:includeArea="includeArea"
```
braughtg: Could be: `v-bind: includeArea="false"`

modules/farm_fd2/src/entrypoints/transplanting/App.vue
```diff
@@ -285,6 +280,7 @@ export default {
       enableSubmit: true,
       enableReset: true,
       rowValues: ['1', '2', '3', '4', '5', '6', '7', '8', '9', '10'],
+      includeArea: false,
```
braughtg: Drop this element because `false` constant was used above.

modules/farm_fd2/src/entrypoints/transplanting/transplanting.soilDisturbance.e2e.cy.js
```diff
@@ -0,0 +1,66 @@
...
+    cy.get('[data-cy="transplanting-soil-disturbance-accordion-title"]')
+      .should('be.visible')
+      .should('contain.text', 'Equipment & Soil Disturbance');
+    cy.get('[data-cy="transplanting-soil-disturbance-accordion-item"]').should(
+      'not.be.visible'
+    );
+  });
+
+  it('Soil disturbance accordion expands to show equipment selector', () => {
+    cy.get(
```
braughtg, line 29: Check if this needs to be clicked twice? If not remove one of the `click` calls.

modules/farm_fd2/src/entrypoints/transplanting/transplanting.submitReset.e2e.cy.js
```diff
@@ -0,0 +1,284 @@
...
+
+    cy.get('[data-cy="soil-disturbance-depth"]')
+      .find('[data-cy="numeric-input"]')
+      .type(9);
+    cy.get('[data-cy="soil-disturbance-depth"]')
+      .find('[data-cy="numeric-input"]')
+      .blur();
+    cy.get('[data-cy="submit-button"]').should('be.enabled');
+
+    cy.get('[data-cy="equipment-selector-1"]')
```
braughtg, line 200: Let's create a separate test that does:
- enter equipment
- make depth invalid
- click submit
- check submit is disabled
- delete equipment
- check submit is enabled

Also add another separate test like this for the speed field.

braughtg (changes requested):
Review of everything except the unit tests.  See comments on the code for requested changes.

modules/farm_fd2/src/entrypoints/transplanting/lib.submitError.unit.cy.js
```diff
@@ -0,0 +1,191 @@
...
+      cy.wrap(
+        lib
+          .submitForm(form)
+          .then(() => {
+            // Shouldn't run because submitForm throws an error.
+            throw new Error('The submission should have failed.');
+          })
+          .catch((error) => {
+            console.log(error.message);
+            expect(error.message).to.contain('Error creating direct seeding.');
```
braughtg: Check error message here and in `lib.js` should be "transplanting"

modules/farm_fd2/src/entrypoints/transplanting/lib.submitError.unit.cy.js
```diff
@@ -0,0 +1,191 @@
...
+            expect(error.message).to.contain(
+              'Result of operation trayInventoryQuantities could not be cleaned up.'
+            );
+            expect(error.message).to.contain(
+              'Result of operation transplantingAsset could not be cleaned up.'
```
braughtg: Consider "transplantedPlantAsset" as name of the operation, here and in `lib.js`.

modules/farm_fd2/src/entrypoints/transplanting/lib.submitError.unit.cy.js
```diff
@@ -0,0 +1,191 @@
...
+    cy.restoreSessionStorage();
+  });
+
+  afterEach(() => {
+    cy.saveLocalStorage();
+    cy.saveSessionStorage();
+  });
+
+  it(
+    'Check error messages when cannot clean up excluding soilDistribution',
```
braughtg: "soilDisturbance"

modules/farm_fd2/src/entrypoints/transplanting/lib.submit.unit.cy.js
```diff
@@ -0,0 +1,410 @@
...
+    cy.restoreSessionStorage();
+  });
+
+  afterEach(() => {
+    cy.saveLocalStorage();
+    cy.saveSessionStorage();
+  });
+
+  it('Check the parents', () => {
+    cy.wrap(farmosUtil.getPlantAsset(result.transplantingAsset.id)).then(
```
braughtg: Consider `transplantedPlantAsset` as operation name here and in `lib.js`.

modules/farm_fd2/src/entrypoints/transplanting/lib.submit.unit.cy.js
```diff
@@ -0,0 +1,410 @@
...
+      (plantAsset) => {
+        expect(plantAsset.relationships.parent[0].id).to.equal(
+          result.parents[0].id
```
braughtg: Tests should check against values in `form` rather than in `result` to verify that the `form` data has been stored in the database and not just that the data sent (i.e. `result`) has been stored.

modules/farm_fd2/src/entrypoints/transplanting/lib.submit.unit.cy.js
```diff
@@ -0,0 +1,410 @@
...
+        );
+        expect(plantAsset.relationships.parent[0].type).to.equal(
+          'asset--plant'
+        );
+      }
+    );
+  });
+
+  it('Check the tray inventory quantity--standard assets', () => {
+    // cy.wrap(farmosUtil.getSeedlings('BROCCOLI')).then((seedlings) => {
```
braughtg: Remove commented out code.

modules/farm_fd2/src/entrypoints/transplanting/lib.submit.unit.cy.js
```diff
@@ -0,0 +1,410 @@
+import { lib } from './lib.js';
+import * as farmosUtil from '@libs/farmosUtil/farmosUtil';
+
+describe('Submission using the transplanting lib.', () => {
+  let form = {
```
braughtg, line 16: Consider creating a new Tray Seeding to be used for this test so it is not dependent upon the current database state.  Can we `import lib.js` from `tray_seeding`?

Add `'@fd2': fileURLToPath(new URL('entry_points/', import.meta.url)),` to the `vite.config.js` and `vite.config.e2e.js` files in `farm_fd2`.

Then import with something like:

`import {lib as traySeedingLib} from @fd2/tray_seeding/lib.js`

modules/farm_fd2/src/entrypoints/transplanting/lib.submit.unit.cy.js
```diff
@@ -0,0 +1,410 @@
...
+    //     seedlings[0].total_trays -
+    //       result.trayInventoryQuantities[0].attributes.value.decimal
+    //   ).to.equal(3);
+    // });
+    cy.wrap(
+      farmosUtil.getStandardQuantity(result.trayInventoryQuantities[0].id)
+    ).then((inventoryQuantity) => {
+      expect(inventoryQuantity.type).to.equal('quantity--standard');
+      expect(inventoryQuantity.attributes.measure).to.equal('count');
+      expect(inventoryQuantity.attributes.value.decimal).to.equal('1');
```
braughtg: Use `form` data when possible. For example, here use `form.picked[0].trays` instead of hard coding `1`.

modules/farm_fd2/src/entrypoints/transplanting/lib.submit.unit.cy.js
```diff
@@ -0,0 +1,410 @@
...
+      expect(transplantingLog.attributes.name).to.equal(
+        '1950-01-02_xp_BROCCOLI'
+      );
+      expect(transplantingLog.attributes.timestamp).to.contain('1950-01-02');
+      expect(transplantingLog.attributes.status).to.equal('done');
+      expect(transplantingLog.attributes.is_movement).to.be.true;
+
+      expect(transplantingLog.relationships.location.length).to.equal(3);
+      expect(transplantingLog.relationships.location[0].id).to.equal(
+        result.transplantingLog.relationships.location[0].id
```
braughtg: Use `form.location` and the `fieldMap` to get the `id` of the field that should have been put into the database.

Similarly for the beds below.

modules/farm_fd2/src/entrypoints/transplanting/lib.submit.unit.cy.js
```diff
@@ -0,0 +1,410 @@
...
+      );
+      expect(transplantingLog.relationships.location[2].id).to.equal(
+        result.transplantingLog.relationships.location[2].id
+      );
+
+      expect(transplantingLog.relationships.asset[0].type).to.equal(
+        'asset--plant'
+      );
+      expect(transplantingLog.relationships.asset[0].id).to.equal(
+        result.transplantingLog.relationships.asset[0].id
```
braughtg: Should this be `result.transplantingAsset.id`?  That is the `asset--plant` created by the transplanting operation.

modules/farm_fd2/src/entrypoints/transplanting/lib.submit.unit.cy.js
```diff
@@ -0,0 +1,410 @@
...
+      expect(transplantingLog.relationships.category[0].type).to.equal(
+        'taxonomy_term--log_category'
+      );
+      expect(transplantingLog.relationships.category[0].id).to.equal(
+        result.transplantingLog.relationships.category[0].id
+      );
+
+      expect(transplantingLog.relationships.quantity.length).to.equal(5);
+
+      for (let i = 0; i < 5; i++) {
```
braughtg: Check these against the quantities in the `result`.

modules/farm_fd2/src/entrypoints/transplanting/lib.submit.unit.cy.js
```diff
@@ -0,0 +1,410 @@
...
+        );
+        expect(speedQuantity.attributes.label).to.equal('Speed');
+
+        expect(speedQuantity.attributes.inventory_adjustment).to.be.null;
+
+        expect(speedQuantity.relationships.units.type).to.equal(
+          'taxonomy_term--unit'
+        );
+        expect(speedQuantity.relationships.units.id).to.equal(
+          result.speedQuantity.relationships.units.id
```
braughtg: Use a map for the `units` here.  Get the map using one of the `farmosUtil.js` functions.

braughtg (changes requested):
Good work.  A few systematic changes. But it's getting close!

modules/farm_fd2/src/entrypoints/transplanting/lib.js
```diff
@@ -106,11 +105,13 @@ export async function submitForm(formData) {
       },
       undo: async (results) => {
         if (results['transplantingLog'] != 'undone') {
-          await farmosUtil.deletePlantAsset(results['transplantingAsset'].id);
+          await farmosUtil.deletePlantAsset(
```
braughtg: I don't think that deleting the transplanting log deletes the plant asset in the same way that it does for quantities.  So I think the plant asset needs to be deleted even if the transplanting log has been undone. But you can check that by running the test and seeing if the plant asset is left behind.  If it is then we don't want the `if` statement here.

modules/farm_fd2/src/entrypoints/transplanting/lib.submit.unit.cy.js
```diff
@@ -0,0 +1,221 @@
...
+    });
+
+    cy.wrap(traySeedingLib.submitForm(traySeedingForm), { timeout: 10000 })
+      .then(() => {
+        return cy.wrap(farmosUtil.getSeedlings(traySeedingForm.cropName), {
+          timeout: 10000,
+        });
+      })
+      .then((res) => {
+        form.picked[0] = { trays: 25, data: res[res.length - 1] };
```
braughtg: Better to use `traySeedingForm.trays` here instead of hard coded `25` in case the form changes.

modules/farm_fd2/src/entrypoints/transplanting/lib.submit.unit.cy.js
```diff
@@ -0,0 +1,221 @@
...
+    expect(result.trayInventoryQuantities[0].attributes.measure).to.equal(
+      'count'
+    );
+    expect(result.trayInventoryQuantities[0].attributes.label).to.equal(
+      'Trays'
+    );
+    expect(
+      result.trayInventoryQuantities[0].attributes.inventory_adjustment
+    ).to.equal('decrement');
+  });
```
braughtg, line 124: We should check the `asset` here to so that we are sure that the inventory of the correct plant asset is being adjusted.  This can be checked against `results.plantAsset.id`

modules/farm_fd2/src/entrypoints/transplanting/lib.submit.unit.cy.js
```diff
@@ -0,0 +1,221 @@
...
+    expect(result.transplantingBedFeetQuantity.type).to.equal(
+      'quantity--standard'
+    );
+    expect(result.transplantingBedFeetQuantity.attributes.measure).to.equal(
+      'length'
+    );
+    expect(result.transplantingBedFeetQuantity.attributes.label).to.equal(
+      'Bed Feet'
+    );
+  });
```
braughtg, line 171: We should check that the inventory adjustment and asset here to be sure that they are not set.  See a similar example in the Direct Seeding unit tests.  Similarly for other quantity tests where quantities are not inventory adjustments.

modules/farm_fd2/src/entrypoints/transplanting/lib.submit.unit.cy.js
```diff
@@ -0,0 +1,221 @@
...
+      'count'
+    );
+    expect(result.trayInventoryQuantities[0].attributes.label).to.equal(
+      'Trays'
+    );
+    expect(
+      result.trayInventoryQuantities[0].attributes.inventory_adjustment
+    ).to.equal('decrement');
+  });
+
```
braughtg, line 125: All of the tests for quantities should also be checking that the units are set correctly.  You can get a map from the unit name to the termID and use that to test that the units are set correctly.

modules/farm_fd2/src/entrypoints/transplanting/lib.submit.unit.cy.js
```diff
@@ -0,0 +1,221 @@
...
+    );
+
+    expect(
+      result.transplantingPlantAsset.relationships.plant_type[0].type
+    ).to.equal(cropToTerm.get(form.cropName).type);
+
+    expect(
+      result.transplantingPlantAsset.relationships.plant_type[0].id
+    ).to.equal(cropToTerm.get(form.cropName).id);
+  });
```
braughtg, line 149: We should check that the original plant asset is a parent of this one.  When a transplanting is made, the plant asset from the tray seeding becomes a parent of the plant asset from the transplanting.

modules/farm_fd2/src/entrypoints/transplanting/lib.submit.unit.cy.js
```diff
@@ -0,0 +1,221 @@
...
+    expect(result.transplantingRowFeetQuantity.type).to.equal(
+      'quantity--standard'
+    );
+    expect(result.transplantingRowFeetQuantity.attributes.measure).to.equal(
+      'length'
+    );
+    expect(result.transplantingRowFeetQuantity.attributes.label).to.equal(
+      'Row Feet'
+    );
+  });
```
braughtg, line 218: `transplantingRowFeetQuantity` is an inventory adjustment, so we should check the inventory adjustment type and the associated inventory asset to be sure we are adjusting the inventory of the correct plant asset.

modules/farm_fd2/src/entrypoints/transplanting/lib.submit.unit.cy.js
```diff
@@ -0,0 +1,221 @@
...
+
+  it('Check the speed quantity--standard', () => {
+    expect(result.speedQuantity.attributes.value.decimal).to.equal(form.speed);
+    expect(result.speedQuantity.type).to.equal('quantity--standard');
+    expect(result.speedQuantity.attributes.measure).to.equal('rate');
+    expect(result.speedQuantity.attributes.label).to.equal('Speed');
+  });
+
+  it('Check the soil disturbance log--activity', () => {
```
braughtg, line 313: Seems like there should be more tests here.  Have a look at the `farmosUtil.js` function for creating a soil disturbance log. Then make sure that each of the parameters passed has an assertion in the test.  That will ensure that the correct values were passed from `lib.js` to the `farmosUtil.js` function.

braughtg (changes requested):
Overall really nice and it's almost there.  Mostly just a few places where we need some additional assertions.  One way to think through what is necessary is for each object created by `lib.js`:
- Look at the documentation for the `farmosUtil.js` function that is used.
- We should have at least one assertion for each parameter that the function accepts. That will ensure that the function was passed the correct values.

PR #297, opened 2024-06-05
Components with add item (+) button clear cached values when item is added

### LINKED ISSUE #103: Components with add item (`+`) button should clear cached values when item is added
A number of components allow new items to be added to the database by clicking a `+` button.  For example the `CropSelector` component allows a new crop to be added and the `LocationSelector` component allows a new location to be added:

image

The values displayed by these components are cached by the `farmosUtil.js` library when they are fetched via the API.  So if a new item (e.g. crop) is added it will not appear in the dropdown when returning to the form.

To observe this:
- Visit the Tray Seeding form
- Click `+` on crop
- Create a new crop "AAA"
- Return to the Tray Seeding form
- The new crop "AAA" does not appear in the list of crops.
- Clear the browser history
- Return to the Tray Seeding form
- The new crop "AAA" now appears in the list of crops.

The fix:
- Have each component with a `+` clear the associated cached items when the `+` is clicked.
  - This will force the component to make an API request for the content the next time it is rendered, and thus the new items will be added to the list.

The components that currently need to be fixed are:
- `components/CropSelector`
- `components/LocationSelector`
- `components/TraySizeSelector`

- Others may have been added since the creation of this issue.  So, review all of the components in the `components` directory to determine if any additional components need to be updated.

Tests should be added to the `*.behavior.comp.cy.js` test file to confirm that clicking the `+` clears the cached items. It is not necessary for these tests to add new items, but they must confirm that the cache has been cleared. Cached items can be cleared using functions in the `library/farmosUtil/farmosUtil.js` library.

Shahir-47:
**Pull Request Description**

Fixes the following components:

- `components/CropSelector`
- `components/LocationSelector`
- `components/TraySizeSelector`

Additionally, adds tests to the `*.behavior.comp.cy.js` test file to confirm that clicking the + clears the cached items. 

closes #103 
---

**Licensing Certification**


braughtg:
When I look at the Files Changed for this PR it has a bunch of commits from an old PR that we merged already.  Did you update `development` before creating this branch or maybe you made it from another feature branch instead of from the most current development?

One way to fix this would be to:
- Update `development`
- Create a new feature branch from `development`
- Switch to the new feature branch
- Checkout just the files that you changed from the branch in this PR into your new feature branch.
  - https://www.freecodecamp.org/news/git-checkout-file-from-another-branch/
- Commit those changes to the new feature branch
- Create a new PR using that branch
- Close this PR.

components/CropSelector/CropSelector.permissions.comp.cy.js
```diff
@@ -9,21 +9,24 @@ describe('Test the permission based CropSelector content', () => {
 
   it('Admin can add crops and url is correct', () => {
     const readySpy = cy.spy().as('readySpy');
+    const addClickedSpy = cy.spy().as('addClickedSpy');
 
     cy.mount(CropSelector, {
       props: {
         includeGreenhouses: true,
         onReady: readySpy,
+        onAddClicked: addClickedSpy,
```
braughtg, line 18: The `CropSelector` should handle the `add-clicked` event emitted from the `Selector` base, clear the cache and go to the new URL.  It will not emit an `add-clicked` event of its own.  So the `addClickedSpy` here doesn't get called.  The tests for the parent components (e.g. `CropSelector`) will need to use the intercept approach that was in `SelectorBase` to be sure that they go to the correct URL when they receive an `add-clicked` event.

components/SelectorBase/SelectorBase.vue
```diff
@@ -206,6 +202,10 @@ export default {
       type: Boolean,
       default: false,
     },
+    includeAddButton: {
```
braughtg, line 211: All props should have a JSDoc comment so that it shows up in the documentation.

PR #305, opened 2024-06-10
Adds popup for + button in SelectorBase

### LINKED ISSUE #152: Return to FD2 form after using + button
Many of the FarmData2 components include `+` buttons to add new farmOS content (e.g. crops, locations, tray sizes, equipment, etc).  When a `+` button is clicked the user is directed to a farmOS form that allows them to create the relevant content.  Currently when the "Save" button is clicked on those forms the user is redirected back to an empty version of the form.  It would be better to have the user redirected back to the FarmData2 form where the `+` was clicked and also to ensure that any form content is retained. 

In addition, if it is possible, the newly created content (e.g. crop, location, etc.) should be selected in the appropriate field under the assumption that it was created so that it could be used.

Related to #103 which will need to be addressed first.

sidlamsal:
**Pull Request Description**

Changes the `+` button in `SelectorBase` to display a popup containing the add asset form.

closes #152 

---

**Licensing Certification**

### DIFF
--- a/components/CropSelector/CropSelector.vue
+++ b/components/CropSelector/CropSelector.vue
@@ -12,7 +12,6 @@
       v-on:update:selected="handleUpdateSelected($event)"
       v-on:valid="handleValid($event)"
       v-on:add-clicked="handleAddClicked"
-      v-bind:includeAddButton="canCreateCrop"
       v-bind:popupUrl="popupUrl"
     />
   </div>
@@ -81,7 +80,7 @@ export default {
     return {
       cropList: [],
       canCreateCrop: false,
-      popupUrl: '/admin/structure/taxonomy/manage/plant_type/add',
+      popupUrl: null,
     };
   },
   methods: {
@@ -100,19 +99,23 @@ export default {
       this.$emit('valid', event);
     },
     async handleAddClicked(newCrop) {
-      // Clear the cached
-      farmosUtil.clearCachedCrops();
-
-      // Populate the map and wait for it to complete
-      await this.populate();
+      // when the selector emits the add-clicked event
+      // clear the cached crops and repopulate the options
+      // to get the newly created crop, then select it
 
       // If a new asset is provided, update the selected
       if (newCrop) {
+        // Clear the cached
+        farmosUtil.clearCachedCrops();
+
+        // Populate the map and wait for it to complete
+        await this.populateCropList();
+
         this.handleUpdateSelected(newCrop);
       }
     },
 
-    async populate() {
+    async populateCropList() {
       try {
         let cropMap = await farmosUtil.getCropNameToTermMap();
 
@@ -132,6 +135,10 @@ export default {
       .then(([canCreate, cropMap]) => {
         this.cropList = Array.from(cropMap.keys());
         this.canCreateCrop = canCreate;
+
+        if (this.canCreateCrop) {
+          this.popupUrl = '/admin/structure/taxonomy/manage/plant_type/add';
+        }
 
         /**
          * The select has been populated with the list of crops and the component is ready to be used.
--- a/components/CropSelector/CropSelector.vue
+++ b/components/CropSelector/CropSelector.vue
@@ -99,8 +99,21 @@ export default {
        */
       this.$emit('valid', event);
     },
-    handleAddClicked() {
+    handleAddClicked(newCrop) {
       farmosUtil.clearCachedCrops();
+      if (newCrop) {
+        this.populate().then(() => {
+          this.handleUpdateSelected(newCrop);
+        });
+      } else {
+        this.populate();
+      }
+    },
+    async populate() {
+      let cropMap = farmosUtil.getCropNameToTermMap();
+      Promise.resolve(cropMap).then((cropMap) => {
+        this.cropList = Array.from(cropMap.keys());
+      });
     },
   },
   watch: {},
--- a/components/CropSelector/CropSelector.vue
+++ b/components/CropSelector/CropSelector.vue
@@ -99,21 +99,28 @@ export default {
        */
       this.$emit('valid', event);
     },
-    handleAddClicked(newCrop) {
+    async handleAddClicked(newCrop) {
+      // Clear the cached
       farmosUtil.clearCachedCrops();
+
+      // Populate the map and wait for it to complete
+      await this.populate();
+
+      // If a new asset is provided, update the selected
       if (newCrop) {
-        this.populate().then(() => {
-          this.handleUpdateSelected(newCrop);
-        });
-      } else {
-        this.populate();
+        this.handleUpdateSelected(newCrop);
       }
     },
+
     async populate() {
-      let cropMap = farmosUtil.getCropNameToTermMap();
-      Promise.resolve(cropMap).then((cropMap) => {
+      try {
+        let cropMap = await farmosUtil.getCropNameToTermMap();
+
+        // Update asset list
         this.cropList = Array.from(cropMap.keys());
-      });
+      } catch (error) {
+        console.error('Error populating crop map:', error);
+      }
     },
   },
   watch: {},
--- a/components/CropSelector/CropSelector.vue
+++ b/components/CropSelector/CropSelector.vue
@@ -13,6 +13,7 @@
       v-on:valid="handleValid($event)"
       v-on:add-clicked="handleAddClicked"
       v-bind:includeAddButton="canCreateCrop"
+      v-bind:popupUrl="popupUrl"
     />
   </div>
 </template>
@@ -80,6 +81,7 @@ export default {
     return {
       cropList: [],
       canCreateCrop: false,
+      popupUrl: '/admin/structure/taxonomy/manage/plant_type/add',
     };
   },
   methods: {
@@ -99,7 +101,6 @@ export default {
     },
     handleAddClicked() {
       farmosUtil.clearCachedCrops();
-      window.location.href = '/admin/structure/taxonomy/manage/plant_type/add';
     },
   },
   watch: {},
--- a/components/SelectorBase/SelectorBase.behavior.comp.cy.js
+++ b/components/SelectorBase/SelectorBase.behavior.comp.cy.js
@@ -155,4 +155,44 @@ describe('Test the SelectorBase behaviors', () => {
         cy.get('[data-cy="selector-popupIframe"]').should('not.exist');
       });
   });
+  it('Clicking add button disables background elements and close enables them', () => {
+    const readySpy = cy.spy().as('readySpy');
+
+    cy.mount(SelectorBase, {
+      props: {
+        required: true,
+        invalidFeedbackText: 'Invalid feedback text.',
+        label: `TheLabel`,
+        options: ['One', 'Two', 'Three', 'Four', 'Five'],
+        onReady: readySpy,
+        popupUrl: '',
+      },
+    });
+
+    cy.get('@readySpy')
+      .should('have.been.calledOnce')
+      .then(() => {
+        cy.get('[data-cy="selector-add-button"]').click();
+
+        cy.get('[data-cy="selector-popupIframe"]')
+          .should('exist')
+          .should('not.have.attr', 'aria-hidden', 'true')
+          .should('not.have.attr', 'tabindex', '-1');
+
+        cy.window().then((window) => {
+          const body = window.document.getElementsByTagName('body')[0];
+          expect(body.getAttribute('tabindex')).to.equal('-1');
+          expect(body.getAttribute('aria-hidden')).to.equal('true');
+        });
+
+        cy.get('[data-cy="selector-closePopup"]').click();
+        cy.get('[data-cy="selector-popupIframe"]').should('not.exist');
+
+        cy.window().then((window) => {
+          const body = window.document.getElementsByTagName('body')[0];
+          expect(body.getAttribute('tabindex')).to.not.equal('-1');
+          expect(body.getAttribute('aria-hidden')).to.not.equal('true');
+        });
+      });
+  });
 });
--- a/components/SelectorBase/SelectorBase.behavior.comp.cy.js
+++ b/components/SelectorBase/SelectorBase.behavior.comp.cy.js
@@ -117,6 +117,7 @@ describe('Test the SelectorBase behaviors', () => {
     const readySpy = cy.spy().as('readySpy');
     const addClickedSpy = cy.spy().as('addClickedSpy');
 
+    // when mounting, set the popupUrl to '' to avoid errors when testing
     cy.mount(SelectorBase, {
       props: {
         required: true,
@@ -158,6 +159,7 @@ describe('Test the SelectorBase behaviors', () => {
   it('Clicking add button disables background elements and close enables them', () => {
     const readySpy = cy.spy().as('readySpy');
 
+    // when mounting, set the popupUrl to '' to avoid errors when testing
     cy.mount(SelectorBase, {
       props: {
         required: true,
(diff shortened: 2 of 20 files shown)


components/SelectorBase/SelectorBase.vue
braughtg, line 1: Be sure to update the JSDoc as appropriate (e.g. update the example from the `SelectorBase` example, once it has been updated.

components/SelectorBase/SelectorBase.vue
```diff
@@ -1,5 +1,36 @@
 <template>
   <div>
+    <div
+      id="overlay"
+      data-cy="overlay"
```
braughtg: Prefix the `id` and `data-cy` for all the new elements with `selector-` to be consistent with other elements. Also add them to the table of `data-cy` attributes in the JSDoc at the start of the `<script>` section.

components/SelectorBase/SelectorBase.vue
braughtg, line 1: The `SelectorBase` example should be updated to use the new `popupUrl` prop.  Then update the JSDoc as appropriate (e.g. update the example from the `SelectorBase` example, and add the new `data-cy` attributes).

components/SelectorBase/SelectorBase.vue
```diff
@@ -212,11 +243,24 @@ export default {
...
+      default: null,
+    },
   },
   data() {
     return {
       optionList: this.options,
       selectedOption: this.selected,
+      isPopupVisible: false,
+      popupSrc: '',
+      isLoaded: false,
```
braughtg: Maybe use `isPopupLoaded` to be more descriptive?

components/SelectorBase/SelectorBase.vue
```diff
@@ -212,11 +243,24 @@ export default {
       type: Boolean,
       default: false,
     },
+    /**
```
braughtg: Does this prop make the `includeAddButton` prop unnecessary? Seems like the condition on the `v-if` for displaying the `+` now should be `popupUrl != null`.  Would that work?

components/SelectorBase/SelectorBase.vue
```diff
@@ -247,6 +291,81 @@ export default {
...
+        const iframeUrl = iframe.contentWindow.location.href;
+        // prohibits navigating away from the add form
+        if (!iframeUrl.includes(this.popupUrl)) {
+          this.hidePopup();
+        }
+      } catch (e) {
+        this.hidePopup();
+      }
+    },
+    getNewAsset(document) {
```
braughtg, line 347: This should have a comment in the method describing what it is doing and why.

components/SelectorBase/SelectorBase.vue
```diff
@@ -247,6 +291,81 @@ export default {
...
+      // when the iframe redirects, hide it
+      iframe.contentWindow.onunload = () => {
+        this.isLoaded = false;
+      };
+
+      // attempts to get & fill-in the newly created asset
+      let result = this.getNewAsset(iframe.contentWindow.document);
+      if (result) {
+        this.hidePopup();
+        this.$emit('add-clicked', result.trim());
```
braughtg: The emit should have a JSdoc comment describing the emitted event and its payload.  See some of the other component for examples.  The docs should be clear about what the payload is if the popup is closed without having created the asset.

components/CropSelector/CropSelector.vue
```diff
@@ -13,6 +13,7 @@
       v-on:valid="handleValid($event)"
       v-on:add-clicked="handleAddClicked"
       v-bind:includeAddButton="canCreateCrop"
+      v-bind:popupUrl="popupUrl"
```
braughtg, line 15: See comment in `SelectorBase` about `popupUrl` and `includeAddButton`.  I think they are redundant.  Here you could use a conditional or a computed property to include the `popupUrl` if `canCreateCrop` is `true` and `null` if it is `false`.

This would apply to all other parent components also.

components/CropSelector/CropSelector.vue
```diff
@@ -97,9 +99,28 @@ export default {
...
+      // Populate the map and wait for it to complete
+      await this.populate();
+
+      // If a new asset is provided, update the selected
+      if (newCrop) {
+        this.handleUpdateSelected(newCrop);
+      }
+    },
+
+    async populate() {
```
braughtg: Can we give this method a more descriptive name?

components/CropSelector/CropSelector.vue
```diff
@@ -97,9 +99,28 @@ export default {
        */
       this.$emit('valid', event);
     },
-    handleAddClicked() {
+    async handleAddClicked(newCrop) {
+      // Clear the cached
```
braughtg: I think one comment describing what this method is doing and why would be more useful than the individual line comments describing what the lines are doing.

components/CropSelector/CropSelector.vue
```diff
@@ -97,9 +99,28 @@ export default {
        */
       this.$emit('valid', event);
     },
-    handleAddClicked() {
+    async handleAddClicked(newCrop) {
```
braughtg, line 101: Could everything in this method be in the if(newCrop) conditional? Seems like if no crop was created then there is no need to clear the cache and reload the map.

If so, this will apply to all of the other parent components also.

braughtg (changes requested):
This is a great feature and it is looking really good! 

I reviewed `SelectorBase.vue`, `CropSelector.vue` and `CropSelector.popup.e2e.cy.js` and left some comments.  Some are local, others suggest changes across the other parent components and tests as well.

components/CropSelector/CropSelector.vue
sidlamsal: I did it that way as a from of error proofing. Currently, ‘add-clicked’ is emitted with the ‘newCrop’ payload only if the ‘SelectorBase’ found the confirmation message and got the name of the new crop (this also closes the popup). If ‘SelectorBase’ cannot find the ‘newCrop’ (add page structure changes or some other reason) and the user closes the popup manually, currently it will still clear the cache and reload the map. Let me know if that’s something that should be implemented or not.

braughtg: It seems like getting `null` back indicates that no new crop would have been created and thus, there is no reason to clear and refresh the cache with a new API call.

components/SelectorBase/SelectorBase.vue
```diff
@@ -312,15 +311,17 @@ export default {
         const iframeUrl = iframe.contentWindow.location.href;
         // prohibits navigating away from the add form
         if (!iframeUrl.includes(this.popupUrl)) {
-          this.hidePopup();
+          this.hidePopup('');
         }
       } catch (e) {
-        this.hidePopup();
+        this.hidePopup('');
```
braughtg: What would result be below (line 325) if the farmOS interface changes and the text for the new item could not be found by that code?  Is it `''` or `null` or `undefined`?  Does it make sense to use that value here so that that case is covered by your code without creating an error?

components/SelectorBase/SelectorBase.content.comp.cy.js
```diff
@@ -144,7 +144,7 @@ describe('Test the default SelectorBase content', () => {
         label: `TheLabel`,
         options: ['One', 'Two', 'Three', 'Four', 'Five'],
         onReady: readySpy,
-        includeAddButton: true,
+        popupUrl: '',
```
braughtg: The empty string (`''`) is kind of special, so passing it here makes it seem at first look like you are doing a test specifically about the empty string.  But it looks like the popup appears as long as `popupUrl` is not `null` - if I am reading things correctly.  So maybe use a non-`null` value here that is not the `''`, just to make it more clear to the reader that you are providing a value here and not specifically testing the behavior of the `''`,

modules/farm_fd2_examples/src/entrypoints/selector_base/App.vue
```diff
@@ -53,14 +53,26 @@
...
-            data-cy="add-button-checkbox"
-            switch
-            v-model="includeAddButton"
-          />
+          <BButton
+            id="popup-url-button"
+            data-cy="popup-url-button"
+            variant="outline-primary"
+            size="sm"
+            v-on:click="this.popupUrl = ''"
```
braughtg: I would suggest using `Not Null Url` or something here and in the button text.  As in my other comment an "Empty URL" or `''` seems somehow special and could be confusing to the reader who (like me) might wonder, "Why an empty URL?" when you just mean a `notNullUrl`

modules/farm_fd2_examples/src/entrypoints/selector_base/App.vue
```diff
@@ -53,14 +53,26 @@
         </td>
       </tr>
       <tr>
-        <td>Include Add Button</td>
+        <td>PopupUrl</td>
```
braughtg, line 56: Seems like only one or the other of these two buttons should be enabled.  Alternatively you could use a switch to add a non-`null` Url instead of two buttons.

braughtg (changes requested):
Overall this is looking really good.  Just a few smaller comments in the code and one larger issue to consider.  

When the popup is visible all of the components in the background are still in the tab order (i.e. can tab through them) and they would be read by a screen reader as if they are usable.  We should figure out how to avoid that.  See the "Accessibility" section for the `Overlay` component for a discussion of this:

https://bootstrap-vue-next.github.io/bootstrap-vue-next/docs/components/overlay.html#accessibility

Once you think you have an idea for how to deal with this post it as a response to this comment and we can discuss it before you spend too much time trying to implement it.

sidlamsal:
So `aria-hidden="true"` will hide the element from screen readers, but it will still be in the tab sequence correct? I tried it out and they were still tab-able. Not sure if I did it wrong. What if we have the `app.vue` bind `:aria-hidden="var"` for the div containing the form elements and `SelectorBase` can emit an event with a payload when the `(+)` button is clicked which the page handles by setting `var` to the payload thus dynamically setting `aria-hidden`. Although, *if* that works as stated, the side bar with the links will still not have `aria-hidden="true"` i think. Maybe a focus-trap which prevents the user from tabbing:
- https://github.com/focus-trap/focus-trap
- https://focus-trap.github.io/focus-trap/

components/SelectorBase/SelectorBase.vue
sidlamsal: It should return `null` since `result=null` before the `try`. So if I'm understanding this correctly, we return the payload `null` here thus signifying that the new crop was not found and then handle that in the parent component?

braughtg: Yes. That way it will be handled consistently for the case of `X` and the case where the added item could not be found.

braughtg:
> So `aria-hidden="true"` will hide the element from screen readers, but it will still be in the tab sequence correct? I tried it out and they were still tab-able. Not sure if I did it wrong. 

I think that is the expected behavior.  Setting `aria-hidden="true"`, ensures that the screen reader doesn't read them.  But I think the elements need to be disabled to be taken out of the tab order.

> What if we have the `app.vue` bind `:aria-hidden="var"` for the div containing the form elements and `SelectorBase` can emit an event with a payload when the `(+)` button is clicked which the page handles by setting `var` to the payload thus dynamically setting `aria-hidden`.

Maybe we should think of having two events `add-clicked` which is emitted when `+` is first clicked and `add-done` which is emitted when the add is finished (either because the item was added or the popup was closed). The parent component could handle the `add-clicked` and `add-done` events.

> Although, _if_ that works as stated, the side bar with the links will still not have `aria-hidden="true"` i think. Maybe a focus-trap which prevents the user from tabbing.

I think you are probably right.  I wonder if there is a way to disable the whole menu?  Maybe setting `disabled="true"` in the parent component for the menu or something?  A `focus-trap` seems like it may be overkill for this situation?

braughtg:
> > Although, _if_ that works as stated, the side bar with the links will still not have `aria-hidden="true"` i think. Maybe a focus-trap which prevents the user from tabbing.
> 
> I think you are probably right. I wonder if there is a way to disable the whole menu? Maybe setting `disabled="true"` in the parent component for the menu or something? A `focus-trap` seems like it may be overkill for this situation?



> > So `aria-hidden="true"` will hide the element from screen readers, but it will still be in the tab sequence correct? I tried it out and they were still tab-able. Not sure if I did it wrong.
> 
> I think that is the expected behavior. Setting `aria-hidden="true"`, ensures that the screen reader doesn't read them. But I think the elements need to be disabled to be taken out of the tab order.
> 
> > What if we have the `app.vue` bind `:aria-hidden="var"` for the div containing the form elements and `SelectorBase` can emit an event with a payload when the `(+)` button is clicked which the page handles by setting `var` to the payload thus dynamically setting `aria-hidden`.
> 
> Maybe we should think of having two events `add-clicked` which is emitted when `+` is first clicked and `add-done` which is emitted when the add is finished (either because the item was added or the popup was closed). The parent component could handle the `add-clicked` and `add-done` events.
> 
> > Although, _if_ that works as stated, the side bar with the links will still not have `aria-hidden="true"` i think. Maybe a focus-trap which prevents the user from tabbing.
> 
> I think you are probably right. I wonder if there is a way to disable the whole menu? Maybe setting `disabled="true"` in the parent component for the menu or something? A `focus-trap` seems like it may be overkill for this situation?

Thinking about this some more... my suggestion is that you see if disabling the elements that make up the menus (either by disabling a top level block, or possibly by iterating over all sub-elements) is possible using JQuery.  If that is not reasonable, then I think we go forward with not disabling them (but still marking the a block containing them hidden).  But then we will want to open a new issue describing the problem.

components/SelectorBase/SelectorBase.behavior.comp.cy.js
```diff
@@ -123,18 +123,76 @@ describe('Test the SelectorBase behaviors', () => {
...
+    const readySpy = cy.spy().as('readySpy');
+
+    cy.mount(SelectorBase, {
+      props: {
+        required: true,
+        invalidFeedbackText: 'Invalid feedback text.',
+        label: `TheLabel`,
+        options: ['One', 'Two', 'Three', 'Four', 'Five'],
+        onReady: readySpy,
+        popupUrl: '',
```
braughtg, line 170: Can this also be a `nonNullValue`?  More generally can any uses of `''` as the URL be changed to reduce changes of confusion?  If there is a reason they cannot just add an explanation.

components/SelectorBase/SelectorBase.vue
```diff
@@ -247,6 +290,117 @@ export default {
...
+    },
+    disableBackgroundElements(disable) {
+      // get the body of the page and remove/restore it from the tab index
+      // and hide/show it for screen readers
+      const element = document.getElementsByTagName('body')[0];
+      if (element) {
+        this.setAttributes(element, disable);
+      }
+    },
+    setAttributes(element, disable) {
```
braughtg, line 317: This seems sensible but I still seem to be tabbing through the elements in the hidden page.  If I click in the date and begin tabbing, eventually the URL of other elements appears at the bottom of the window when I continue to tab.  

Is that the same behavior your are experiencing?

components/SelectorBase/SelectorBase.behavior.comp.cy.js
sidlamsal: It seems to cause an error during testing when ever the (+) is clicked and the iframe has its source set. I commented out all the code i added which tampers with the iframe (checking URL, removing elements, etc.) and even with a plain iframe it caused errors. Since these tests do not test for the content of the iframe (that is tested the e2e tests) i think empty string is probably sensible solution.

components/SelectorBase/SelectorBase.vue
sidlamsal: For me, the only tab-able elements are in the search bar (new tab, back button, bookmark etc.) and the background html tag element is also tab able. None of the background components or links were tab-able. I checked this by removing the code that re-enables the elements and closing the popup and seeing what else i could tab to.

PR #315, opened 2024-06-17
Add multi crop selector entry point

### LINKED ISSUE #312: Build a `MultiCropSelector` component
Create a component for selecting multiple crops (i.e. a seed mixture as for cover crops).

This component should appear approximately as follows:
image

Notes:
- This should behave like the `EquipmentSelector` component.
  - Initially only one blank entry appears.
  - If a crop is selected, another blank entry appears below it.
  - The last entry includes a `+` if the user has permission to create plant assets.
  - All entries but the last one include a "trash" icon to delete them.
- Use the `EquipmentSelector` as a model for the props and events to be implemented.

Building this component should also include:
- An example page in farm_fd2_examples for the component.
- A set of component tests for the component.

sidlamsal:
**Pull Request Description**

Adds:
- multi crop selector component
- component tests
- example page

closes #312 

---

**Licensing Certification**


sidlamsal:
Still working on making the component look exactly like the drawn example.

braughtg:
> Still working on making the component look exactly like the drawn example.

The drawn example was a good guess, but if you have alternatives you'd like to discuss, post them in a comment on the issue.  Otherwise, I'll look for you to flip this to a full PR and then I'll have a look at it.

components/MultiCropSelector/MultiCropSelector.vue
```diff
@@ -0,0 +1,193 @@
...
+     * Whether validity styling should appear on input elements.
+     */
+    showValidityStyling: {
+      type: Boolean,
+      default: false,
+    },
+  },
+  data() {
+    return {
+      selectedCrop: this.selected,
```
braughtg: This should probably be `selectedCrops` to make it clear there can be multiple.

components/MultiCropSelector/MultiCropSelector.adminPermissions.comp.cy.js
```diff
@@ -0,0 +1,36 @@
+import MultiCropSelector from '@comps/MultiCropSelector/MultiCropSelector.vue';
```
braughtg: Do we actually have to click the `+` button here, or is it enough to just check that the `+` button is visible?  If so, then this test and the `guestPermissions` tests can be combined into one test file.

The `behavior` test will be the one that has to click the button and ensure that it goes to the correct page (or eventually that the correct page is loaded into the popup).

braughtg (changes requested):
I only had time for a quick review of this so far.  A few things that I noticed:

- The crop selector drop down should expand to fill the available area to its right.  See the `SelectorBase` example or the way the `EquipmentSelector` behaves in the `SoilDisturbance` example.
- The `+` should use a popup.  But I think this will largely take care of itself when #305 is merged. May just need a few prop changes.

modules/farm_fd2_examples/src/module/farm_fd2_examples.links.menu.yml
```diff
@@ -58,3 +58,9 @@ farm.fd2_examples_example_entry_point:
   parent: farm.fd2_examples_main
   route_name: farm.fd2_examples_example_entry_point.content
   weight: 100
+farm.fd2_examples_multi_crop_selector:
+  title: 'MultiCropSelector Example'
```
braughtg: Change title to be just `MultiCropSelector` here so that the word "Example" does not appear in the menu.

braughtg (changes requested):
One small change here.  Then I'd like to do a full in person review tomorrow or Friday.

PR #318, opened 2024-06-20
feat(fd2): adds entry point for soil disturbance events

### LINKED ISSUE #249: Add Soil Disturbance Entry Point
Add a new entry point to `farm_fd2` for Soil Disturbances.

- This entry point should:
  - be created using `addEntrypoint.bash`:
    - be named `soil_disturbance`
    - have the title `Soil Disturbance` 
    - appear on the "Soil" submenu on the FarmData2 menu
    - have an appropriate permission string
      - For example: `access content,create plant asset,create seeding log,create standard quantity`
        - Note the above is an example, but is not correct for this entry point.
      - The available permissions can be found by using dev-tools to inspect the checkboxes at:
        - http://farmos/admin/people/permissions
- It will likely include (or be entirely?) a `SoilDisturbance` component.
- A set of associated unit tests for the entry point's `lib.js` file.
- A set of e2e tests for the entry point

The data to be collected must be based on the PASA Soil Health Benchmark Study information below:
image

sidlamsal:
**Pull Request Description**

Adds:
- soil disturbance entry point
- lib.js including tests
- e2e tests

Closes #249 

---

**Licensing Certification**

Co-authored-by: Shahir Ahmed [98346408+Shahir-47@users.noreply.github.com](mailto:98346408+Shahir-47@users.noreply.github.com)

### DIFF
--- a/library/farmosUtil/farmosUtil.checkPlantAsset.unit.cy.js
+++ b/library/farmosUtil/farmosUtil.checkPlantAsset.unit.cy.js
@@ -86,14 +86,14 @@ describe('Test the bed utility functions', () => {
   });
 
   it('Get plant assets in trays only', () => {
-    const locationName = 'JASMINE';
-    const locIds = [greenHouseMap.get('JASMINE').id];
+    const locationName = 'CHUAU';
+    const locIds = [greenHouseMap.get('CHUAU').id];
 
     cy.wrap(farmosUtil.getPlantAssets(locationName, [], true, false)).then(
       (plantAssetIds) => {
         plantAssetIds.forEach((id) => {
           cy.wrap(farmosUtil.getPlantAsset(id)).then((asset) => {
-            expect(asset.attributes.inventory).to.equal(true);
+            expect(asset.attributes.inventory[0].units).to.equal('TRAYS');
             const plantLocIds = asset.relationships.location.map(
               (loc) => loc.id
             );
@@ -108,14 +108,14 @@ describe('Test the bed utility functions', () => {
   });
 
   it('Get plant assets in ground only', () => {
-    const locationName = 'JASMINE';
-    const locIds = [greenHouseMap.get('JASMINE').id];
+    const locationName = 'CHUAU';
+    const locIds = [greenHouseMap.get('CHUAU').id];
 
     cy.wrap(farmosUtil.getPlantAssets(locationName, [], false, true)).then(
       (plantAssetIds) => {
         plantAssetIds.forEach((id) => {
           cy.wrap(farmosUtil.getPlantAsset(id)).then((asset) => {
-            expect(asset.attributes.inventory).to.not.equal(true);
+            expect(asset.attributes.inventory[0].units).to.not.equal('TRAYS');
             const plantLocIds = asset.relationships.location.map(
               (loc) => loc.id
             );
--- a/library/farmosUtil/farmosUtil.checkPlantAsset.unit.cy.js
+++ b/library/farmosUtil/farmosUtil.checkPlantAsset.unit.cy.js
@@ -153,7 +153,7 @@ describe('Test the getPlantAssets function', () => {
   });
 
   it('Get plant assets with no matching beds', () => {
-    const locationName = 'JASMINE';
+    const locationName = 'CHUAU';
     const checkedBeds = ['NonExistentBed1', 'NonExistentBed2'];
 
     cy.wrap(farmosUtil.getPlantAssets(locationName, checkedBeds)).then(
--- a/library/farmosUtil/farmosUtil.checkPlantAsset.unit.cy.js
+++ b/library/farmosUtil/farmosUtil.checkPlantAsset.unit.cy.js
@@ -1,6 +1,6 @@
 import * as farmosUtil from './farmosUtil.js';
 
-describe('Test the bed utility functions', () => {
+describe('Test the getPlantAssets function', () => {
   let bedMap = null;
   let fieldMap = null;
   let greenHouseMap = null;
@@ -33,6 +33,7 @@ describe('Test the bed utility functions', () => {
 
     cy.wrap(farmosUtil.getPlantAssets(locationName, checkedBeds)).then(
       (plantAssetIds) => {
+        expect(plantAssetIds).to.have.length.greaterThan(0);
         plantAssetIds.forEach((id) => {
           cy.wrap(farmosUtil.getPlantAsset(id)).then((asset) => {
             expect(asset.attributes.status).to.equal('active');
@@ -53,6 +54,7 @@ describe('Test the bed utility functions', () => {
 
     cy.wrap(farmosUtil.getPlantAssets(locationName, checkedBeds)).then(
       (plantAssetIds) => {
+        expect(plantAssetIds).to.have.length.greaterThan(0);
         plantAssetIds.forEach((id) => {
           cy.wrap(farmosUtil.getPlantAsset(id)).then((asset) => {
             const plantLocIds = asset.relationships.location.map(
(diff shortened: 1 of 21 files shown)


library/farmosUtil/farmosUtil.js
```diff
@@ -1644,6 +1644,103 @@ export async function getPlantAsset(plantAssetId) {
...
+    if (plantAsset.attributes.status !== 'active') {
+      continue;
+    }
+
+    let addToResults = false;
+
+    if (isInTrays && isInGround) {
+      addToResults = true;
+    } else if (isInTrays && !isInGround) {
+      if (plantAsset.attributes.inventory[0].units === 'TRAYS') {
```
braughtg: Use the `farmosUtil.getAssetInventory` function here instead of assuming the `TRAYS` value will always be at index 0.

library/farmosUtil/farmosUtil.js
```diff
@@ -1644,6 +1644,103 @@ export async function getPlantAsset(plantAssetId) {
...
+    locationAsset = greenhouseAsset;
+  } else if (fieldAsset) {
+    locationAsset = fieldAsset;
+  } else {
+    return [];
+  }
+
+  // Fetch all plant assets
+  const plantAssets = await farm.asset.fetch({
+    filter: { type: 'asset--plant' },
```
braughtg: Can some of the filtering can be done on the server side by adding properties to the `filter` here? In particular, can we filter for `active` assets and assets that are in the location and possibly beds we are interested in?   Then do the filtering for `inTrays` or `inGround` in the browser (if it can't be done server side).

Fetching all of the `asset--plant` objects could take a while on an active farm with possibly 100's or 1000's of active plants (and event more archived ones from past years.)

You can find more information about the `filter` object here: https://github.com/farmOS/farmOS.js/blob/main/docs/entities.md

braughtg (changes requested):
I have only looked at the `getPlantAssets` function so far.  There are a few suggestions in the code.

library/farmosUtil/farmosUtil.js
sidlamsal: filtering for active plants worked, however filtering for location as such,
```
  const plantAssets = await farm.asset.fetch({
    filter: {
      type: 'asset--plant',
      status: 'active',
      'location.id': locationAsset.id,
    },
  });
```
gives,
 
```XHRGET
http://farmos/api/asset/plant?filter[status-0-filter][condition][path]=status&filter[status-0-filter][condition]
[operator]==&filter[status-0-filter][condition][value]=active&filter[location.id-0-filter][condition]
[path]=location.id&filter[location.id-0-filter][condition][operator]==&filter[location.id-0-filter][condition]
[value]=9854eb1c-d215-4385-8989-40773ee7d9db2
[HTTP/1.1 500 Internal Server Error 93ms]
	
jsonapi	Object { version: "1.0", meta: {…} }
errors	[ {…} ]
0	Object { title: "Internal Server Error", status: "500", detail: "Call to a member function getColumns() on bool", … }
title	"Internal Server Error"
status	"500"
detail	"Call to a member function getColumns() on bool"
links	Object { via: {…}, info: {…} }
```

When I tried to use the example filter used in the guide `'owner.id': '22222222-2222-2222-2222-222222222222'`, it ran without error. Additionally filtering for `plant_type.id` and `parent.id` both ran without error. However `inventory.unit` also had the same error as `location.id`.

braughtg: I think this is probably related to this farmos.js but that still hanging around: https://github.com/farmOS/farmOS.js/issues/86

I'll try to have a go at fixing that farmos.js bug this week.  Then we can use a custom version or maybe they will merge a PR for us.

For now, we can leave this as it is for testing, but once it's merged we'll want to create a ticket in the issue tracker that describes the problem so that we don't lose track of it.

library/farmosUtil/farmosUtil.checkPlantAsset.unit.cy.js
```diff
@@ -0,0 +1,160 @@
...
+    cy.saveLocalStorage();
+    cy.saveSessionStorage();
+  });
+
+  it('Check that returned plant assets are active', () => {
+    const locationName = 'ALF';
+    const checkedBeds = ['ALF-1', 'ALF-2'];
+
+    cy.wrap(farmosUtil.getPlantAssets(locationName, checkedBeds)).then(
+      (plantAssetIds) => {
```
braughtg, line 35: Note that this test will pass if the call to `getPlantAssets` fails and returns an empty array.  A similar thing happens in the other tests as well.  Maybe add a check that the result is not empty, and then check that all the assets are active?

modules/farm_fd2/src/entrypoints/soil_disturbance/App.vue
```diff
@@ -0,0 +1,391 @@
...
+          <template v-slot:label>
+            <span
+              id="termination-event-label"
+              data-cy="termination-event-label"
+              class="p-0"
+              >Termination Event:</span
+            >
+          </template>
+
+          <BFormCheckbox
```
braughtg, line 82: I'm not seeing the termination event check box at all when I use the Soil Disturbance form.  Not sure why that would be.

modules/farm_fd2/src/entrypoints/soil_disturbance/soil_disturbance.equipment.e2e.cy.js
braughtg, line 1: The tests here should also check that the equipment is required.

library/farmosUtil/farmosUtil.js
```diff
@@ -1644,6 +1644,117 @@ export async function getPlantAsset(plantAssetId) {
...
+  if (greenhouseAsset) {
+    locationAsset = greenhouseAsset;
+  } else if (fieldAsset) {
+    locationAsset = fieldAsset;
+  } else {
+    return [];
+  }
+
+  // Fetch all plant assets
+  const plantAssets = await farm.asset.fetch({
```
braughtg, line 1707: Add `limit: Infinity` to this filter so that we get all of the relevant plants.  Otherwise, we will only get one page.  See `getLogCategories` or something similar for an example.

braughtg (changes requested):
Getting really close.  See the embedded comments for a few issues.

modules/farm_fd2/src/entrypoints/soil_disturbance/App.vue
braughtg: Was due to browser caching.  Problem resolved when the browser cache was cleared.

braughtg (approved):

braughtg:
Some prior tests change the database in ways that can make this test fail.  So we just need to ensure that the database is reset before the `soil_disturbance/submission.e2e.cy.js` test is run.  You can do this by adding the following `before` hook in that test:

```
  before(() => {
    cy.task('initDB');
  });
```

braughtg:
When manually testing the Soil Disturbance endpoint I noticed a few things:
- The comment is not being added to the Activity Logs that are created.  The comment provided should be added to the log for each pass.
- The Area (%) is not updated based upon the beds that are selected.  For example, if 2/4 beds are selected then the Area (%) should default to 50% of the area and allow the user to adapt it from there.

One other small improvement that I think we should make is:
- when there are multiple passes, add "Pass 1." or "Pass 2.", etc to the beginning of the comment for each log.

The tests should be also be augmented to check for these things.

library/farmosUtil/farmosUtil.js
braughtg, line 1: Looks good.  Since this updates the `farmosUtil.js` it should also update at least one of the tests in `farmosUtil.soil.unit.cy.js` to include and check the comment.

modules/farm_fd2/src/entrypoints/soil_disturbance/lib.js
braughtg, line 1: Nice! I like the including the total # of passes here too.

modules/farm_fd2/src/entrypoints/soil_disturbance/lib.submit.unit.cy.js
```diff
@@ -143,7 +143,7 @@ function runTest(terminationValue) {
       );
     });
 
-    Cypress._.times(2, (i) => {
+    Cypress._.times(form.passes, (i) => {
```
braughtg, line 146: Good catch here!!

braughtg (changes requested):
Everything looks good.  Just one request for a small test change since `farmosUtil.js` was also modified here.

braughtg (approved):

PR #327, opened 2024-06-24
Restyles the MultCropSelector component

### LINKED ISSUE #323: Restyle the MultCropSelector component
The component currently looks like:

image

It would look better to style this more similar to the way the equipment selector is displayed:

image

So we should move the "Crops" heading to a box that looks similar to the accordion header for the soil disturbance component in the Direct Seeding form.  This box should not be an accordion but the heading should look like that same box without the `V` drop down indicator.  This will be similar to what was done for the Equipment heading on the Soil Disturbance form.

### LINKED ISSUE #329: MultiCropSelector: `valid` event not emitted correctly when required and last equipment is deleted
## Observed Behavior

On the MultiCropSelector example page:
- Ensure that the component is required.
- Select a crop and observe that `valid` event is emitted with `true`
- Click the "Remove Last" button
- Notice that the `valid` event box still indicates `true` 

## Desired Behavior

When a crop is required and the list becomes empty, a `valid` event should be emitted with the value `false`.

## Also

A test case should be written for this.  Ideally, the test should be written first so it can be observed as failing.  Then correct the bug, then fix the test and observe it passing.

Related to #319, #320, #321 and all may be able to be solved with a single PR.

Shahir-47:
**Pull Request Description**

The component looked like: 

image

This pull request changes it to the style similar to the way the equipment selector is displayed:

image

Closes #323 
Closes #329 

---

**Licensing Certification**

### DIFF
--- a/components/MultiCropSelector/MultiCropSelector.vue
+++ b/components/MultiCropSelector/MultiCropSelector.vue
@@ -1,34 +1,35 @@
 <template>
-  <div
-    id="multi-crop-selector"
-    data-cy="multi-crop-selector"
-  >
-    <label
-      id="multi-crop-selector-label"
-      data-cy="multi-crop-selector-label"
-      for="crop-selector-container"
-      >Crop(s)
-    </label>
+  <div>
     <div
-      id="crop-selector-container"
-      data-cy="crop-selector-container"
+      id="multi-crop-selector"
+      data-cy="multi-crop-selector"
     >
-      <SelectorBase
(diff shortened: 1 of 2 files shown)


braughtg (changes requested):
The restyling looks good.  Well done.

Just a few things to do:
- update the docs with all of the `data-cy` elements that appear in the page.
- double check the static example in the docs against the example and update if necessary.
- add a link in the docs to the live example (see other components with an example for a sample).

braughtg:
Superseded by #332 which adds a test.

Issue #319, opened 2024-06-21
EquipmentSelector should only require one piece of equipment


braughtg:
The `EquipmentSelector` component allows users to select multiple pieces of equipment used in an activity (e.g. a Direct Seeding or Soil Disturbance).  When equipment in an `EquipmentSelector` is `required`, only the first piece of equipment should be indicated as required.

## To Observe
- Visit FD2 Examples -> Components -> Soil Disturbance
- Select at least 3 pieces of equipment

![image](https://github.com/FarmData2/FarmData2/assets/16962482/8dc4746f-c4d0-4c3f-aa38-61ef1570465a)

Notice that:
- the first two pieces of equipment include the red `*` indicating that they are required.
- only the third piece of equipment includes a delete icon (trash can).

This makes it inconvenient to remove the first or second piece of equipment if desired.

## Desired Behavior

- When an equipment selector is required, only the first selector should require a value (i.e. it is the only one that should ever have the `*` indicating that it is required.
- If a second piece of equipment is selected then:
  - The first selector should retain its `*` but should have a delete icon.
  - The second piece of equipment should not have a `*` and should have a delete icon.
- Similarly if more than two pieces of equipment are selected.
  - The first selector should retain its `*` but should have a delete icon.
  - All other pieces of equipment should not have `*` and should have a delete icon.
- If pieces of equipment are deleted others below them should shift up.
  - If only one piece of equipment remains, it should not have a delete icon.

Note this behaivor is similar to what happens when the `EquipmentSelector` is not required.

## Also

- As a part of this issue, an example page should be made for the `EquipmentSelector` component (see #169). 
- The component tests for the `EquipmentSelector` should be updated to test this behavior.

braughtg:
Closed by #338

PR #338, opened 2024-06-29
Changes required indicator and delete button behavior for EquipmentSelector and MultiCropSelector

### LINKED ISSUE #320: MultiCropSelector should require only one crop
The `MultiCropSelector` component allows users to select multiple crops (e.g. in a Cover Crop Seeding).  When a crop in an `MultiCropSelector` is `required`, only the first crop should be indicated as required.

## To Observe
- Visit FD2 Examples -> Components -> MultiCropSelector
- Select at least 3 crops

image

Notice that:
- the first two crops include the red `*` indicating that they are required.
- only the third crop includes a delete icon (trash can).

This makes it inconvenient to remove the first or second crops if desired.

## Desired Behavior

- When an a multicrop selector is required, only the first selector should require a value (i.e. it is the only one that should ever have the `*` indicating that it is required.
- If a second crop is selected then:
  - The first selector should retain its `*` but should have a delete icon.
  - The second crop should not have a `*` and should have a delete icon.
- Similarly if more than two crops are selected.
  - The first selector should retain its `*` but should have a delete icon.
  - All other crops should not have `*` and should have a delete icon.
- If crops are deleted others below them should shift up.
  - If only one crop remains, it should not have a delete icon.

Note this behavior is similar to what happens when the `MultiCropSelector` is not required.

## Also

- The component tests for the `MultiCropSelector` should be updated to test this behavior.

sidlamsal:
**Pull Request Description**

Adds the `includeDeleteButton` prop to `SelectorBase` to handle including both a delete button and the required `*`. 

Updates `MultiCropSelector` and `EquipmentSelector` so that when multiple items are selected, only the top item has `*` and all items have the delete button. With only one selected item, the top item has `*` but no delete button.

Additional testing for the 3 components have been added as well as an updated `SelectorBase` example page. 

closes #320 #319 

---

**Licensing Certification**

### DIFF
--- a/components/EquipmentSelector/EquipmentSelector.content.comp.cy.js
+++ b/components/EquipmentSelector/EquipmentSelector.content.comp.cy.js
@@ -99,6 +99,62 @@ describe('Test the default EquipmentSelector content', () => {
       });
   });
 
+  it('Check required prop', () => {
+    const readySpy = cy.spy().as('readySpy');
+
+    cy.mount(EquipmentSelector, {
+      props: {
+        required: true,
+        onReady: readySpy,
+        selected: [],
+      },
+    });
+
+    cy.get('@readySpy')
+      .should('have.been.calledOnce')
+      .then(() => {
+        // no items => required and no delete
+        cy.get('[data-cy="equipment-selector-1"]')
+          .find('[data-cy="selector-delete-button"]')
+          .should('not.exist');
+        cy.get('[data-cy="equipment-selector-1"]')
+          .find('[data-cy="selector-required"]')
+          .should('exist');
+        // one item => required and no delete
+        cy.get('[data-cy="equipment-selector-1"]')
+          .find('[data-cy="selector-input"]')
+          .select('Tractor');
+        cy.get('[data-cy="equipment-selector-1"]')
+          .find('[data-cy="selector-delete-button"]')
+          .should('not.exist');
+        cy.get('[data-cy="equipment-selector-1"]')
+          .find('[data-cy="selector-required"]')
+          .should('exist');
+        cy.get('[data-cy="equipment-selector-2"]')
+          .find('[data-cy="selector-delete-button"]')
+          .should('not.exist');
+        cy.get('[data-cy="equipment-selector-2"]')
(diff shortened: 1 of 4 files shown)


braughtg (changes requested):
This looks pretty good.  

I find the interaction between the `includeDeleteButton` and `required` properties of `SelectorBase` in this PR to be a little confusing.

The way the `SelectorBase` worked before this PR is that the delete button appears when the `SelectorBase` is not required and something is selected.  Otherwise the delete button does not appear.

So it seems like, just changing the `isRequired` function in `MultiCropSelector` and `EquipmentSelector` (as you have done) so that only the first element is required would be sufficient.

If so then the`SelectorBase` component would not need to be changed at all.  The changes to `EqupimentSelector` and `MultiCropSelector` would be just to the `isRequired` method (because the `includeDeleteButton` prop goes away).

Do you think that will work?

sidlamsal:
> Do you think that will work?

The reason I added the new prop is because `required` was used in the `v-if` of the delete button and the indicator. As per the issue #320, when a second crop is added, the first crop should have the indicator and a delete button. I could not find a way to make this happen (by only changing `required(i)`) as even if only the first crop is required, it would not have a delete button when another crop is selected. My original changes were actually a little smaller, `includeDeleteButton` had a default value of `!this.required` so it was less changes. However this caused an issue when running the exists e2e test for the multi crop example page when it would try to read the value `this.required` and would throw the error `Cannot read properties of undefined (reading 'required')`. We could just remove the `!required` from the `v-if` for the delete button, but I wasn't sure if that would be the desired behavior, so I tried to implement the new behavior with minimal changes to old behavior.

braughtg:
> We could just remove the `!required` from the `v-if` for the delete button

I think this is a good idea... maybe make the first one required only when there is 0 or 1 crop/equipment selected.  So once a second crop or equipment is selected then none of them are required (i.e. any of them can be deleted).  But once we get down to 1 crop or equipment selected then it will be required again.  Seems like that should be doable without changes to the `SelectorBase`.

braughtg:
Nicely done! I like this fix!

PR #339, opened 2024-07-01
fix(comp): factors out MultiSelectorBase component

### LINKED ISSUE #321: Refactor EquipementSelector and MultiCropSelector to extract parent component
The `EquipmentSelector` and `MultiCropSelector` components have common behavior.  It may be possible to refactor them by extracting a `MultiSelectorBase` component that captures this common behavior and is then used by both of them.

Addressing this issue should include:
- Creating the parent component.
- Adding an example for the parent component.
- Refactoring the `EquipmentSelector` component.
- Refactoring the `MultiCropSelector` component.
- Updating and adding tests as necessary.

This issue can subsume #319, #320, #328, and #329 if it is completed before them.

sidlamsal:
**Pull Request Description**

Adds `MultiCropSelector` component.
Adds example page.
Adds tests.
Refactors `MultiCropSelector`.
Refactors `EquipmentSelector`.
Adjusts tests where necessary.

Closes #321 

---

**Licensing Certification**


components/MultiSelectorBase/MultiSelectorBase.vue
```diff
@@ -0,0 +1,191 @@
...
+      (valid) => {
+        validity.selected = valid;
+      }
+    "
+    v-on:ready="createdCount++"
+    v-on:error="(msg) => showErrorToast('Network Error', msg)"
+  />
+ * ```
+ *
+ * ## `data-cy` Attributes
```
braughtg, line 57: The docs here should be updated.

modules/farm_fd2_examples/src/entrypoints/multi_selector_base/App.vue
braughtg, line 1: I'm seeing some strange behavior related to the "Select first option" button in this example.  I am not sure if it is caused by the example page or by the component itself. 

To see one example:
1. Open the example page
2. Click "Select first option" three times
    - There should be 3 "One"s in the list now
3. Delete the first item
    - Now I see an empty first and third item and "One" in the second item.

braughtg (changes requested):
As a first pass I manually tested the `MultiSelectorBase` using the example page and found a little bit of strange behavior.  It looks like the original `EquipmentSelector` has the same issue so I suspect it came from there. The comment on the `SelectorBase.vue` file gives some more details.

components/MultiCropSelector/MultiCropSelector.vue
```diff
@@ -73,11 +70,11 @@ import * as farmosUtil from '@libs/farmosUtil/farmosUtil.js';
  * multi-crop-selector         | The container div for the MultiCropSelector component.
```
braughtg: Should be `multi-crop-selector-main` here I think.

braughtg (changes requested):
There should be a `popup` test in the `MultiSelectorBase` for the `+` button to ensure that it works in the base component.   With that in place, the `popup` tests for the `MultiCropSelector` and `EquipmentSelector` can be simplified.

The `behavior` test for the `EquipmentSelector` and the `MultiCropSelector` can be simplified as they do not need to test the behavior of the `MultiSelectorBase` (e.g. that selecting one thing adds another selector, etc).  That will all be covered by the tests for the `MultiSelectorBase`.

braughtg (changes requested):
These changes all look really good.  Unfortunately, they also broke a few of the tests in the `SoilDisturbance` component because it uses the `EquipmentSelector`.  

Please run `test.bash --comp --glob=components/SoilDisturbance/*.comp.cy.js` and address the failing tests and this will be good to merge.  This should just reqiure adjusting the tests to fit the new `MultiSelectorBase` behaviors rather than any changes to the `SoilDisturbance` component.

PR #366, opened 2024-07-11
fix(dev):  does not require `--unit` flag when running `test.bash --lib` command

### LINKED ISSUE #142: `test.bash` Improvements
- [x] #375
- [ ] fix(dev): `test.bash`: run `e2e.cy.js` tests associated with components
  - Some components include both component (`comp`) and end-to-end (`e2e`) tests (to test them in a context when necessary).  So running the `--comp` tests should run both the `comp` and the `e2e` tests.
  - It should be checked / ensured that this change also changes the tests that are run by `lintstaged` in the git commit hook.
- [ ] fix(dev): `runAllTests.bash`: update command for `--comp` tests 
  - When the task just above is completed the `runAllTests.bash` script should be updated to use the new command for running the `--comp` tests. 
- [ ] fix(dev): `test.bash` should build the modules that are under test. 
  - When `test.bash` is run it should build whatever module is under test (e.g. `fd2`, `examples`, `school`).  This will ensure that any changes to the actual module (i.e. files in the `module` directory) are incorporated before the tests are run.  This is particularly important when merging changes from development that change the module into a feature branch, because if the module is not rebuilt the tests will run against an old version of the module from the `dist` directory.

Comment on this issue to request to work on one of these tasks.  A new issue will be created for the task and will be assigned to you.

### LINKED ISSUE #375: fix(dev): do not require  flag when running  tests.
All `lib` tests are unit tests so it is redundant to specify `--lib --unit`.  The script should be modified so that `test.bash --lib` is a valid command that run all of the unit tests in the `library` directory (as `test.bash --lib --unit` does now`.

Shahir-47:
**Pull Request Description**

The `test.bash` is modified so that `test.bash --lib` is a valid command that run all of the unit tests in the `library` directory (as `test.bash --lib --unit` does now`.

Closes #375 
---

**Licensing Certification**

### DIFF
--- a/bin/test.bash
+++ b/bin/test.bash
@@ -11,6 +11,7 @@ function usage {
   echo "    -e|--e2e  : Run end-to-end tests."
   echo "    -c|--comp : Run component tests."
   echo "    -u|--unit : Run unit tests."
+  echo "    -b|--lib  : Run all unit tests in the library directory."
   echo ""
   echo "  -i|--gui  : Include to run tests within the Cypress GUI test runner."
   echo "              NOTE: if -i|--gui is used with -p|--prev or -l|--live then"
--- a/bin/test.bash
+++ b/bin/test.bash
@@ -126,6 +126,7 @@ while true; do
     ;;
   -b | --lib)
     TEST_LIB=1
+    UNIT_TESTS=1
     shift 2
     ;;
   -g | --glob)


braughtg:
Modules can now contain unit tests (e.g. for `lib.js`) which requires `--unit --fd2` (etc), so I think at this point continuing to require `--unit --lib` is fine.

PR #385, opened 2024-09-30
Refactor PicklistBase to Use Map for Selected Rows and Fix Sorting Issues

### LINKED ISSUE #383: Refactor PicklistBase Component: Improve `picked` Array to Store Selected Rows by `uuid` or Index
## Issue Description

We need to remodel the `PicklistBase` component so that the `picked` array only stores the selected rows (as `uuid`s or original indices), rather than using `1`s and `0`s to track selections. This will resolve an issue where sorting causes a mismatch between the displayed rows and the actual data.

### Why This is an Issue
Currently, when sorting is applied via the column headers, the `sortedRows` (an internal copy of the `rows` array) is re-ordered to reflect the sort order. However, the `picked` array, which tracks selected items with `1`s and `0`s, is also shuffled to align with the sorted `sortedRows` data. 

Because the `rows` array (the original unsorted data) remains unchanged, there is a discrepancy between the sorted `picked` array and the original data. This means that when a user selects a row in the sorted list, it may map to the wrong item in the unsorted `rows` array. 

This problem becomes more significant when the `PicklistBase` component is used by other pages (such as Transplanting), as these pages expect the `picked` array to match the original order of the `rows` data. Without this match, selections may reference incorrect data entries, leading to errors in submission or other processes.

### Why It Needs Refactoring
This issue highlights the need for `picked` to store references to the selected rows using either the `uuid` of each item or the original index in the `rows` array. By doing so, we avoid the confusion caused by reordering the internal `sortedRows` array. This refactor will ensure that sorting behavior does not interfere with the correctness of selections and that developers using the `PicklistBase` component can work with it in a predictable manner without needing to manipulate or access its internal state (e.g., `sortedRows`). 

The refactor will:
1. Change `picked` to only store selected rows, identified by their `uuid` or original index in the `rows` array.
2. Ensure sorting via the `handleSort` method continues to work, while preserving the correct relationship between selected items and the original data.
3. Prevent users from needing to understand the internal structure of `sortedRows`, allowing them to use the `PicklistBase` component without worrying about the internal sorting logic.

#### Component Updates
- **PicklistBase.vue**: Refactor the `picked` array to only store selected rows (either by original index or `uuid`).

#### Component Tests Updates
- `PicklistBase.behavior.comp.cy.js`
- `PicklistBase.content.comp.cy.js`
- `PicklistBase.event.comp.cy.js`
- `PicklistBase.behavior.styling.cy.js`

#### TransplantingPicklist Component & Tests Updates
Since `TransplantingPicklist` uses `PicklistBase`, the following files will also need to be updated:
- **TransplantingPicklist.vue** (or equivalent component)
- `TransplantingPicklist.styling.comp.cy.js`
- `TransplantingPicklist.events.comp.cy.js`
- `TransplantingPicklist.content.comp.cy.js`
- `TransplantingPicklist.behavior.comp.cy.js`

#### Transplanting Page and Tests Updates
The `Transplanting` page uses the `TransplantingPicklist` component, which in turn uses `PicklistBase`. The following files will need to be updated:
- **Transplanting.vue** (or equivalent component)
- `transplanting.crop.e2e.cy.js`
- `transplanting.submission.e2e.cy.js`
- `transplanting.submitReset.e2e.cy.js`

#### Example Pages Updates
The following example pages need to be updated to reflect the changes in `PicklistBase`:
- **modules/farm_fd2_examples/src/entrypoints/picklist_base/App.vue**
- **modules/farm_fd2_examples/src/entrypoints/transplanting_picklist/App.vue**

#### End-to-End Test Updates
The end-to-end tests related to transplanting also need to be updated:
- **modules/farm_fd2/src/entrypoints/transplanting/transplanting.*.e2e.cy.js**

#### Library File Updates
- **lib.js**: Relevant changes need to be applied to support the `PicklistBase` updates as used in Transplanting.

### Relation to Issue #372
This restructuring of the `PicklistBase` component will also help address the requirements of [#372](https://github.com/YOUR_REPO_NAME/issues/372) — *Transplanting: manage selection by crop x bed combinations*. By refactoring the `PicklistBase` component to handle individual plant x bed combinations correctly, we ensure that selections are accurately tracked and mapped, allowing for proper submission events as described in the original issue.

## Summary of Changes

This refactor will:
- Ensure the `picked` array only tracks selected rows by `uuid` or original index.
- Resolve issues caused by sorting by ensuring selections always map to the correct data in the `rows` array.
- Ensure developers can use `PicklistBase.vue` without needing to understand or manipulate its internal `sortedRows` structure.

Shahir-47:
**Pull Request Description**

This PR refactors the `PicklistBase.vue` component to store selected rows in a `Map`, resolving the sorting-related issues described in #383. The internal `pickedRows` array is kept for `BFormSelect` compatibility, while the external `picked` prop now uses a `Map` for cleaner selection handling. Updates include:
- `PicklistBase.vue` refactor
- Changes to `TransplantingPicklist.vue`
- Updates to example pages and associated tests
- Modifications to end-to-end transplanting tests

Closes #383 

---

**Licensing Certification**


components/PicklistBase/PicklistBase.vue
```diff
@@ -576,32 +591,50 @@ export default {
...
+        const pickedMap = new Map();
+        for (let i = 0; i < this.pickedRows.length; i++) {
+          if (this.pickedRows[i] > 0) {
+            // Only add non-zero selections
+            const row = this.sortedRows[i];
+            const originalIndex = this.rows.indexOf(row);
+            if (originalIndex !== -1) {
+              pickedMap.set(originalIndex, {
+                row,
+                quantity: this.pickedRows[i],
```
braughtg: Let's change this property to be `picked`.  That will make it less likely to be confused with the `quantityAttribute` that is used to specify the number of things that are available to be picked.  This will likely have a knock-on effect on `TransplantingPicklist`, but shouldn't go any further than that.

Also, the `update:picked` event needs have some JSDoc describing the event and its payload.  See the `valid` event above for an example.

components/PicklistBase/PicklistBase.vue
```diff
@@ -304,15 +304,15 @@ export default {
       required: true,
     },
     /**
-     * An array of values indicating the rows/quantities that have been picked in the table.
-     * The rows are indexed from 0.
-     * A non-zero value indicates that the row is picked and the quantity if `units` prop is set.
-     * A zero value indicates that the row is not picked.
-     * The length of this array must be equal to the length of the array provided by the `rows` prop.
+     * A Map indicating the rows/quantities that have been picked in the table.
+     * The keys of the Map are the original row indices from the `rows` array.
```
braughtg: `The keys of the Map are row indices from the `rows` array.`

I don't think `original` matters here since the `rows` array is not actually changed.

components/PicklistBase/PicklistBase.vue
```diff
@@ -304,15 +304,15 @@ export default {
...
+     * The values of the Map are objects containing information about the picked row, including the quantity if the `units` prop is set.
+     * If the row is picked, the corresponding entry in the Map will have a `quantity` greater than 0.
```
braughtg: If the row is picked, the Map will have an entry with the row number as the key and the value will have a `quantity` attribute containing a `1` if the `units` prop is not set, or the value picked if the `units` prop is set.

components/PicklistBase/PicklistBase.vue
```diff
@@ -304,15 +304,15 @@ export default {
...
+     * The Map can be empty if no rows are picked.
```
braughtg: `The Map will be empty if no rows are picked.`

modules/farm_fd2_examples/src/entrypoints/picklist_base/App.vue
```diff
@@ -10,7 +10,7 @@
     id="picklist"
     data-cy="picklist"
     v-bind:required="required"
-    invalidFeedbackText="At least one row must be selected."
+    v-bind:invalidFeedbackText="'At least one row must be selected.'"
```
braughtg: Not sure I understand this change...  Is the `v-bind` necessary when the value is a constant?

braughtg (changes requested):
Overall really nice work!  

Just a few small changes and it should be ready to go.

Shahir-47:
@braughtg This PR is ready for review

braughtg:
Well done!

PR #386, opened 2024-10-02
NumericInput: Button Increment Rounding

### LINKED ISSUE #204: NumericInput: Round to nearest increment when using increment buttons
The value in a `NumericInput` is equal to the smallest increment then when the medium or large increments are clicked the value in the field should be rounded down to the value of that increment.  

Examples (assuming the smallest increment is 1):
- If the value in the field is 1 and the large increment is 100, if the large increment button is clicked, the value in the field should become 100, not 101.
- If the value in the field is 1 and the medium increment is 10, if the medium increment button is clicked, the value in the field should become 10, not 11.

RolandLocke:
**Pull Request Description**

Added:
- Form input now follows below rules to help users better understand the behavior:
```
If the component is in its initial state (i.e. has not been manually edited and no increment or decrement buttons have been clicked):
  If an increment button larger than the current value is clicked (e.g. value 60, click +100)
    The value is set to the next higher value of that increment (e.g. value -> 100).
  otherwise (e.g. value 60, click +10)
    The value is incremented based on the button that is clicked. (e.g. value -> 70)
otherwise
  The value is incremented or decremented based on the button clicked.
```
- Button disabling and styling when decrementing the input will drop it below the min value. 
- The input form now has `initialValueChanged` data, indicating when the user has first interacted with the form input.

Closes #204 

---

**Licensing Certification**

FarmData2 is a [Free Cultural Work](https://freedomdefined.org/Definition) and all accepted contributions are licensed as described in the LICENSE.md file. This requires that the contributor holds the rights to do so. By submitting this pull request **I certify
/ that I satisfy the terms of the [Developer Certificate of Origin](https://developercertificate.org/)** for its contents.

### DIFF
--- a/components/NumericInput/NumericInput.behavior.comp.cy.js
+++ b/components/NumericInput/NumericInput.behavior.comp.cy.js
@@ -68,7 +68,7 @@ describe('Test the NumericInput component behavior', () => {
       props: {
         label: 'Test',
         invalidFeedbackText: 'Test feedback text',
-        value: 1,
+        value: 0,
         incDecValues: [1, 10, 100],
         onReady: readySpy,
       },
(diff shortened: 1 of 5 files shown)


braughtg:
Had a look at this this morning and noticed one thing with regard to the minimum value.  When playing with this on the examples page, with the minValue to 0, the small decrement button is disabled (correct). But if I then click the small increment button making the value 1, the small decrement button is still not enabled (incorrect).  If I click the small increment button again, making the value 2, then the small decrement button is enabled.  So there seems to be an off by one error in there somewhere.  Similar behavior occurs when the minValue is not 0.  I did not test the medium or large buttons.

braughtg:
As we discussed in person yesterday, you were seeing an issue where the small increment button was not becoming enabled as expected when the value was increased when running in a component test.  However, the behavior was correct when looking at the `NumericInput` component on the example page.  

I have figured out what the issue is and pushed some changes to this branch.  Briefly, the issue is that the changes you had made to the component rely on `this.value`.  When used in a page the value `prop` is `v-model`ed to a `data` element in the page.  This way, when the component emits the `update:value` event, the value goes to the page and then comes back to the component via the `v-model`.  The new value is then seen inside of the component and all works as expected. 

However, when running in a component test, there is no `v-model` that feeds the new value back into the component. So the component is only seeing the initial value of the `value` `prop` that is set in the test.  I.e. changing the value in the text field does emit the `update:value` event, but the page does not in turn update the associated `prop`.  So when the component looks at the `value` `prop` it is still seeing the initial value. 

The fix was to create an internal numeric value in `data` that can then be used to test things numerically against the current value of the field.   Check the diff for the commit that I pushed for the details.

RolandLocke:
> The fix was to create an internal numeric value in `data` that can then be used to test things numerically against the current value of the field. Check the diff for the commit that I pushed for the details.

I moved the assignment of this variable to the `valueAsString()` method as it applies to text and form inputs in addition to inc/dec button presses.

RolandLocke:
@braughtg I noticed that the rounded corners on the buttons weren't being rendered on the rightside. 

Before:
image

After:
image

This was due to a `class="has-validation"` class assignment missing from the `inputGroup` element. Discussion [linked](https://github.com/bootstrap-vue/bootstrap-vue/issues/5314). I fixed it on this page, but it looks like its an issue on almost every input with a button attached:
image

braughtg:
> This was due to a `class="has-validation"` class assignment missing from the `inputGroup` element. Discussion [linked](https://github.com/bootstrap-vue/bootstrap-vue/issues/5314). I fixed it on this page, but it looks like its an issue on almost every input with a button attached:

Good find and fix.  Would you be willing to open an issue and create a PR to fix this in the `SelectorBase` component as well?

components/NumericInput/NumericInput.vue
braughtg, line 1: There seems to be a disagreement between naming and behavior in the functions used for disabling the increment buttons.  For example, the `numeric-increase-small` button is enabled/disabled using the `diableLargeInc` computed property which doesn't make semantic sense:
```
v-bind:variant="disableLargeInc ? 'outline-secondary' : 'outline-success'
```

It appears that this works because the `disableLargeInc` property uses `incDecValues[0]` which is actually the small increment/decrement value.  

A similar issue exists in the other functions used for the medium and large increment buttons.

components/NumericInput/NumericInput.behavior.comp.cy.js
```diff
@@ -61,6 +61,166 @@ describe('Test the NumericInput component behavior', () => {
...
+
+        cy.get('[data-cy="numeric-decrease-lg"]').click();
+
+        cy.get('[data-cy="numeric-increase-sm"]').should('not.be.disabled');
+        cy.get('[data-cy="numeric-increase-md"]').should('not.be.disabled');
+        cy.get('[data-cy="numeric-increase-lg"]').should('not.be.disabled');
+      });
+  });
+
+  it('Input of a larger value replaces the initial value', () => {
```
braughtg: This test and the next one have the same `it` message.  They should be edited to be descriptive of what they are specifically testing, without getting too long.

braughtg: Also seems like there should be a similar test for the medium increase button.

braughtg (changes requested):
Looking good.  A few comments describing some improvements that should be made.

In addition, I added a comment to the issue asking about the behavior of the decrement buttons on the first click.  Should they behave similarly to the way the increment buttons behave?

components/NumericInput/NumericInput.behavior.comp.cy.js
```diff
@@ -61,6 +61,166 @@ describe('Test the NumericInput component behavior', () => {
...
+
+    cy.get('@readySpy')
+      .should('have.been.calledOnce')
+      .then(() => {
+        cy.get('[data-cy="numeric-increase-lg"]').click();
+        cy.get('[data-cy="numeric-input"]').should('have.value', '100');
+      });
+  });
+
+  it('Input of a larger value replaces the initial value', () => {
```
braughtg: Should there also be a test for the small button for when there are decimals in the value?  For example, min value is 0, current value is 0.5 and the small increment is 1.

braughtg (changes requested):

RolandLocke:
> Good find and fix. Would you be willing to open an issue and create a PR to fix this in the `SelectorBase` component as well?
 
See new [pull request ](https://github.com/FarmData2/FarmData2/pull/392)+ [issue](https://github.com/FarmData2/FarmData2/issues/391)

components/NumericInput/NumericInput.behavior.comp.cy.js
RolandLocke: I've updated both these tests to have a single increment value either larger or smaller than the initial value. This tests all cases as the logic to determine if the initial value should be replaced all happens in the the same function.  It simply depends on the `incDecValues[]` defined by the form.

components/NumericInput/NumericInput.behavior.comp.cy.js
RolandLocke: I'm unsure if it is necessary, but I've added a test for a decimal value just for safety.

braughtg:
I pushed a few changes to the `SoilDisturbance` tests because they relied on clicking buttons that are now disabled.

braughtg:
Just pushed some changes to the `TraySeeding` tests to address similar issues.

PR #388, opened 2024-10-09
feat(fd2): adds plant x bed-specific soil disturbance functionality and termination option

### LINKED ISSUE #372: Soil Disturbance: should allow selection of crop x bed affected
The Soil Disturbance form currently has a checkbox for "Termination Event," which appears only if the location/beds selected contain active plant assets.  When this box is checked all plants in the location/beds are archived.

In practice an area or bed can contain many different plant assets and it is a common occurrence to terminate just one or a few of those plants, but not all of them.  In addition, a plant may be in multiple beds, and only one or a few, but not all of the beds could be terminated.

Thus, the "Termination Event" indication must be done on a plant x bed basis.  To do this, the "Termination Event" checkbox should be expanded into a `PicklistBase` element that displays the list of plants in the selected location/beds.  There should be one entry for each plant x bed (i.e. if a plant is in two beds, then it appears twice in the list).  Each row should display the crop name, the bed name and the date it was planted in the ground (direct seeded or transplanted).  If a location does not have beds associated with it then the plant will be listed only once. Like the "Termination Event" checkbox, this `PicklistBase` should only appear if there are plants in the selected location/beds.

Currently the "Termination Event" checkbox is only displayed if a call to `farmosUtil.getPlantAssets` returns some records.  Those records contain sufficient information to populate the plant x beds list.  All rows of of this list should be selected by default.

If a plant x bed row is selected in a submission, the bed should be removed from the plant's locations.  If the plant has no remaining locations, then the plant should be archived.

The e2e test for the "Soil Disturbance" entry point will need to be updated.

Shahir-47:
### **Pull Request Description**

This pull request addresses the functionality requested in issue [#372](https://github.com/FarmData2/FarmData2/issues/372), which focuses on expanding the Soil Disturbance entry point to allow plant x bed-specific soil disturbance events, including termination options.

The current implementation included a single "Termination Event" checkbox that would terminate all crops in a location. This approach did not support selective actions, especially in cases where only certain plants or beds need to be affected. This PR implements the following features to resolve this limitation:

#### **1. Enhancements to Soil Disturbance:**
- **Termination Event Checkbox Expansion:**
  - Replaced the single "Termination Event" checkbox with a **PicklistBase** table for greater granularity.
  - The picklist allows users to select specific plant x bed combinations for soil disturbance. Each row represents a combination of a crop and planting date. The bed column is shown only if the location has beds.
  - If the location has no plants, the **bed picker** ensures all available beds are selected by default, and the picker replaces the table to allow bed-specific selection.

- **Dynamic Bed Selection Behavior:**
  - The page automatically selects all the beds in the bed picker by default if a location has no plants. This ensures seamless operation for locations without active plant assets.

- **Enhanced Soil Disturbance Options:**
  - Users can now choose specific plant x bed combinations for soil disturbance, offering more control compared to the previous all-or-nothing approach.
  - Users can optionally mark the "Terminate" checkbox to indicate a termination event.

- **New Soil Disturbance Logging Logic:**
  - When termination is indicated, the associated plant x bed rows are logged, and affected beds are removed from the plant’s locations.
  - If a plant has no remaining beds, it is automatically archived.

#### **2. Backend Changes (via Merge of PR #401):**
- Incorporated functionality from PR [#401](https://github.com/FarmData2/FarmData2/pull/401), which introduced the following:
  - **`createSoilDisturbanceTerminationLog`:** Logs termination events and archives plants with no remaining beds.
  - **`getSoilDisturbanceTerminationLog` and `deleteSoilDisturbanceTerminationLog`:** Handle retrieval and deletion of termination logs.
  - Updated **unit tests** for these functions to ensure robust backend behavior.

#### **3. Component Updates:**
- **`LocationSelector.vue`:**
  - Added an event to emit the total available beds for a location to the parent component, ensuring dynamic updates when locations with no active plants are selected.

- **`App.vue` (Soil Disturbance Entry Point):**
  - Added the **Termination Event Checkbox**, **Termination Event Picklist**, and **Bed Picker**.
  - Updated data fields to support the new components.
  - Modified area field calculations to account for the updated soil disturbance behavior.
  - Ensured proper integration of the bed picker when locations with no plants are selected.

#### **4. Testing Updates:**
Comprehensive testing was implemented to cover all scenarios introduced in this PR:

- **Backend Tests:**
  - Created a new test file: `farmosUtil.soilDisturbanceTerminationLog.unit.cy.js` to test the following:
    - Partial bed termination.
    - Full bed termination and automatic plant archiving.
    - Assets without beds and error handling.

- **Component Tests:**
  - Added a test in `LocationSelector.events.comp.cy.js` to verify that the new event emitting total available beds functions correctly.

- **End-to-End Tests for Soil Disturbance:**
  - Updated `soil_disturbance.equipment.e2e.cy.js` to ensure area field behavior adjusts to the new termination picklist.
  - Added a new test: `soil_disturbance.termination.e2e.cy.js` to verify:
    - Termination picklist behavior.
    - Proper handling of scenarios with and without plant assets.
  - Modified `soil_disturbance.submission.e2e.cy.js` and `soil_disturbance.submitReset.e2e.cy.js`:
    - These tests now handle locations with and without plant assets.
    - Validated the new soil disturbance behavior.
    - Ensured proper form reset behavior after a submission.

- **`lib.js` Testing:**
  - Created `lib.submit.unit.cy.js` to test all possible user actions for submissions, including combinations of plant assets, selected beds, and soil disturbance.
  - Created `lib.submitError.unit.cy.js` to test proper error handling and rollback behavior when network issues occur.

#### **5. Default Behavior Adjustments:**
- If no active plantings are present:
  - All available beds are selected by default using the bed picker.
- If active plantings are present:
  - The termination picklist rows are deselected by default, allowing users to make deliberate choices.

---

**Closes #372**

---

### **Licensing Certification**


Shahir-47:
@braughtg The PR is ready for review. I’ve updated the description to reflect all changes. PR #401 was merged into this, so I’ll be closing it. Let me know if you have any questions!

library/farmosUtil/farmosUtil.soilDisturbanceTerminationLog.unit.cy.js
```diff
@@ -0,0 +1,596 @@
...
+    });
+
+    // Delete all logs and plant asset
+    cleanupLogsAndAsset({
+      deleteSoilDisturbanceLog: true,
+      deletePlantAsset: true,
+    });
+  });
+
+  it(
```
braughtg: I think that just one error test for creating the log is sufficient.  They all intercept the same endpoint and that endpoint is accessed by the same code in all cases.  I suggest keeping and renaming this test and removing the other error tests for the log.  Keep the one for the error deleting the log though.

library/farmosUtil/farmosUtil.js
```diff
@@ -2383,6 +2383,131 @@ export async function getTraySeededCropNames() {
...
 }
 
+/**
+ * Creates an activity log (`log--activity`) for a soil disturbance termination event.
+ *
+ * This function performs two key actions:
+ * 1. Terminates specific beds associated with a plant asset at a given location.
+ * 2. Archives the plant asset **if no beds remain after the termination process**.
+ *
+ * **Workflow:**
```
braughtg: You can remove the workflow heading and steps from this comment.

components/LocationSelector/LocationSelector.vue
```diff
@@ -102,7 +102,14 @@ import { BAccordion } from 'bootstrap-vue-next';
 export default {
   name: 'LocationSelector',
   components: { SelectorBase, BedPicker, BAccordion },
-  emits: ['error', 'ready', 'update:selected', 'update:beds', 'valid'],
+  emits: [
+    'error',
+    'ready',
+    'update:selected',
+    'update:beds',
+    'update:allBeds',
```
braughtg: Instead of emitting this new event, have the `LocationSelector` internally set the beds that are selected by default. To do this:
- add a `prop` `selectAllBedsByDefault` which defaults to `false`.
  - when this `prop` is `true` all beds will be selected when the `BedPicker` is first shown.
    - You should be able to use the `beds` computed property to initialize the `checkedBeds` array. 
  - when this `prop` is `false` no beds will be selected when the `BedPicker` is first shown.

This should then also be tested in the `LocationSelector` tests and the tests for the `allBeds` event should be removed.

modules/farm_fd2/src/entrypoints/soil_disturbance/lib.submit.unit.cy.js
```diff
@@ -1,42 +1,34 @@
...
+function runTest(activePlantAsset, terminationValue) {
+  describe(`Test the Soil Disturbance lib submission with activePlantAssets=${activePlantAsset} termination=${terminationValue}`, () => {
+    // Skip the test if the configuration is invalid
+    if (!activePlantAsset && terminationValue) {
+      // Can't terminate without an active plant asset involved
+      it('Should catch invalid configuration and mark as complete', () => {
+        cy.log(
+          'Invalid configuration caught: Termination cannot be true without an active plant asset. Marking this test as complete.'
+        );
+        expect(true).to.be.true; // Mark the test as passing
```
braughtg: Seems like this is saying that this function should never be called with `activePlantAsset=false` and `terminationValue=true`.  So I think the thing to do here is to have this test fail in this situation.  Then remove your call to it at the bottom of the file.  That way if someone in the future make the call with (`false`, `true`) their test will immediately fail.  But otherwise, this will have no effect on the tests.

braughtg (changes requested):
This is shaping up well.  I reviewed the changes in:
- `LocationSelector`
- `farmosUtil.js` & `farmosUtil.soilDisturbanceTests.unit.cy.js`

I did not review the changes to the `soil_disturbance` entrypoint.  However, with regard to that you may find it useful to use the following snippet in tests where there is an undesirable dependency between test files where changes made in one test file cause tests in another test file to fail.  In such a case, adding this snippet to the test file with the failing test case will cause the database to be reset before running the tests in that file.

```Javascript
  before(() => {
    cy.task('initDB');
  });
```

Resetting the database takes some time, so this should only be done when necessary.

Shahir-47:
Hi @braughtg,

I encountered an issue where directly selecting checkboxes from `LocationSelector.vue` caused the `handleUpdateBeds` method in `LocationSelector` to be called twice:  

1. Once from the event triggered and emitted by this line:  
   ```vue
   v-on:update:selected="handleUpdateSelected($event)"
   ```  
2. A second time if I manually called `handleUpdateBeds(this.bedList)` to select all beds.  

To avoid this duplication, I introduced a `selectAll` prop in `PickerBase.vue`, which triggers the `pickAll()` method to check all boxes when needed.  

In `BedPicker.vue`, I set `selectAll` to `true` by watching the `location` prop. If the location changes and the `selectAllBedsByDefault` prop is `true`, it triggers the selection.  

Finally, in `LocationSelector.vue`, I declared the `selectAllBedsByDefault` prop and passed it down to `BedPicker.vue`.  

I've also updated three example pages for testing and exploration:  
- `modules/farm_fd2_examples/src/entrypoints/picker_base/App.vue`  
- `modules/farm_fd2_examples/src/entrypoints/bed_picker/App.vue`  
- `modules/farm_fd2_examples/src/entrypoints/location_selector/App.vue`  

The updated implementation now works correctly on the Soil Disturbance page:  
- `modules/farm_fd2/src/entrypoints/soil_disturbance/App.vue`  

If this approach looks good, I will proceed with writing tests for each of these components. Let me know if you have any feedback or suggestions!

Shahir-47:
Apart from the implementation details I described earlier regarding selecting all beds by default when there are no active plantings, I’ve made all the requested changes based on your review comments.  

The other endpoints, aside from `LocationSelector.vue` and its related child components, are now ready.

braughtg:
It seems to me that the issue with reactivity in `BedPicker` and `PickerBase` can be resolved by improving the behavior of those components rather than adding the new `prop` all the way down.  I took a pass that the `BedPicker` to confirm that this seems like a viable approach.  You can see those changes in the `pickerbase-update` branch (https://github.com/FarmData2/FarmData2/tree/pickerbase-update).  

The basic idea was to emit the `update:picked` event not when `picked` changes, but when the internal `data.checked` value changes.  This also simplified the logic a little by relying on Vue reactivity instead of explicit calls to `updatePicked()` when `data.checked` is changed internally. 

I think  that `BedPicker` will work as is with this change. Then `LocationSelector` can be updated in a way similar to what was done in `PickerBase` so that to select or clear the picked beds it modifies an internal `data` property. So `LocationSelector` changes `data.checkedBeds` that will propagate up to `BedPicker` and then to `PickerBase` which will emit the `update:picked` event which will propagate back down to `BedPicker` and then to `LocationSelector` which will propagate it back to its parent page.

One note: The change in the `picker-update` branch breaks a test for `LocationSelector` related to clearing the beds when the location changes.  That will likely be fixed by making the corresponding changes to the `LocationSelector` so that it works with the modified `PickerBase`.

I think that this PR is getting unmanageablely large.  I propose that we break this off into its own issue/PR, get it working and then merge it to development and then merge that into your feature branch for the plant x bed changes.  To that end, I have created a new issue describing what this new PR should contain.  See #406.

Shahir-47:
@braughtg All changes related to `LocationSelector.vue` have been removed from this branch. Only changes to `farmosUtil` and `soil_disturbance` remain.  

The changes related to `LocationSelector.vue` are now in this PR: [FarmData2 PR #413](https://github.com/FarmData2/FarmData2/pull/413).

Shahir-47:
@braughtg I've completed testing the soil disturbance functionality. All lib.js unit tests, `farmosUtil` soil disturbance termination log tests, and the majority of soil disturbance e2e tests are passing successfully.

However, there's one failing test case related to bed selection when transitioning between locations. The test fails when switching from a location with active plant assets to a location without active plant assets, where beds should be selected by default.

The root cause is in the interaction between `SoilDisturbance.vue` and `LocationSelector`:

```vue
<LocationSelector
  v-bind:allowBedSelection="!plantsAtLocation"
  v-bind:selectAllBedsByDefault="true"
/>
```

The `allowBedSelection` prop depends on `plantsAtLocation`:
```javascript
plantsAtLocation() {
  return this.form.affectedPlants.length > 0;
}
```

When a location change occurs, `LocationSelector`'s beds computed property executes:
```javascript
beds() {
  if (!this.allowBedSelection) {
    return [];
  }
  // ... bed logic
}
```

The failure sequence is:
1. User starts at location with active plant assets (`allowBedSelection` is false)
2. User switches to location without active plant assets
3. `LocationSelector` computes beds with stale `allowBedSelection` value
4. `beds` computed property returns empty array as `allowBedSelection` is false
5. `selectAllBedsByDefault` has no effect due to empty `beds` array
6. `affectedPlants` updates after checking the location
7. `allowBedSelection` becomes true, but bed selection opportunity has passed

A solution would be to emit the bed array to `SoilDisturbance` and handle selection after the `affectedPlants` update completes.

Shahir-47:
@braughtg,

I added the watcher on `allowBedSelection` prop in `LocationSelector` as suggested so all beds are checked automatically. Now, the `soil_disturbance.termination.e2e.cy.js` test should pass, and as a result, all the tests (including the one in this PR) are now passing.

I couldn't test the `allowBedSelection` prop as a component test because Cypress doesn't re-render components on prop changes. I found this potential solution to that problem: [#17712 (comment)](https://github.com/cypress-io/cypress/issues/17712#issuecomment-1646614336). However, as of now, it seems only Cypress React has this feature: [#15984](https://github.com/cypress-io/cypress/issues/15984).

Fortunately, the `soil_disturbance.termination.e2e.cy.js` verifies that all the boxes are checked when switching from `allowBedSelection=false` to `allowBedSelection=true` (and vice versa) when toggling between a location with active plant assets and one without. Apart from the small testing limitation I mentioned, everything else works as expected.

braughtg:
When I `runAllTests.bash` I am seeing one failing e2e test.

```
Running:  soil_disturbance/soil_disturbance.equipment.e2e.cy.js                         (27 of 55)
  1 failing
  1) Soil Disturbance: Equipment Component
       Updating beds updates area:
      Timed out retrying after 10000ms
      + expected - actual
      -'40'
      +'67'
      
      at Context.eval (http://farmos/__cypress/tests?p=src/entrypoints/soil_disturbance/soil_disturbance.equipment.e2e.cy.js:53:137)
```

@Shahir-47 can you look into that one?

modules/farm_fd2/src/entrypoints/soil_disturbance/App.vue
```diff
@@ -243,18 +318,28 @@ export default {
...
+      } else {
+        this.form.area = 100;
+      }
+    },
+    handlePickedUpdate(picked) {
+      this.form.picked = picked;
+
+      if (this.form.affectedPlants.length > 0 && this.form.picked.size > 0) {
+        this.form.area = Math.round(
+          (this.form.picked.size / this.form.affectedPlants.length) * 100
```
braughtg: I'm not sure that this area estimate is correct.  I think we want the number beds affected by the picked items divided by the total number of beds in the area.  For example, if we pick Chuau as the location and then select a plant in Chuau-1 then the result should be 20% because there are 5 beds in Chuau and we are affecting one of them.

Do you agree?

modules/farm_fd2/src/entrypoints/soil_disturbance/lib.submitError.unit.cy.js
```diff
@@ -1,154 +1,330 @@
...
+              .then(() => {
+                // Shouldn't run because submitForm throws an error.
+                throw new Error('The submission should have failed.');
+              })
+              .catch((error) => {
+                expect(error.message).to.contain(
+                  'Error creating Soil Disturbance records.'
+                );
+                expect(error.message).to.contain(
+                  'Result of operation terminationLog0 could not be cleaned up.'
```
braughtg, line 215: Do you recall the reason why these can logs and quantities cannot be cleaned up?

modules/farm_fd2/src/entrypoints/soil_disturbance/lib.submitError.unit.cy.js
```diff
@@ -1,154 +1,330 @@
...
-            expect(plantAssetPatch).to.equal(4);
-            expect(standardQuantityDeletes).to.equal(6);
-            expect(activityLogDeletes).to.equal(1);
-          }),
-        { timeout: 10000 }
-      );
-    }
-  );
-});
+runTest(false);
```
braughtg, line 330: What is the reason that `runTest(false)` is called twice?

braughtg (changes requested):

braughtg:
In reviewing this I am finding that assigning the log category `termination` to the log that removes beds from the plant asset to be a little strange.  This log is labeled as a `termination` log, but then lists the beds that were not terminated. It has to have this specific list of beds because it is a movement log that changes the locations for the plant asset to be the the beds where the plant is still located.  But it is confusing because it feels like it should list the beds that were terminated, because it is a `termination` log.  

Suggestion for discussion:
- Retain the `termination` log category.
- Add a comment indicating that the log removing terminated beds from the plant asset and list the removed beds.
  - e.g. `Removes bed ALF-1 from 2019-02-05_RADISH`
  - e.g. `Removes beds ALF-1, ALF-2 from 2019-02-05_RADISH.`

Tests should be modified (or added if necessary) to check that the comment is correct when:
- One bed is removed.
- Multiple beds are removed.

braughtg:
Currently if a plant asset has one (or more) of it's beds terminated there is a movement log created. The purpose of this log is to remove the terminated beds from the locations of the plant asset.  This works well.  However, if a plant asset with no beds associated with it is terminated, this log is still created. In this case, that log does not serve any purpose because the location does not change (i.e. no beds are removed).  So, if a plant asset with no beds associated with it is terminated, this movement log should not be created.  A test should be added for this case.

Shahir-47:
**@braughtg**  

With the latest update, if a plant asset has no beds, a movement log is no longer created. Previously, if the movement log failed, the archival would also not take place because it came after the log creation step. However, now that the movement log **never runs** for plant assets without beds, we can’t rely on its failure to prevent the plant from being archived when an error occurs.  

For locations that have active plant assets but no beds, the **area will be 100%**.

Additionally, if a bed is **partially chosen**—for example, if there are two crops in `ALF-1` and only one is selected—then `ALF-1` **will not count towards the area percentage**. A bed will only contribute to the percentage if **all** crops within it are selected.

Lastly, I **updated the Soil Disturbance Termination Log tests** and the **related Soil Disturbance tests** to account for these new additions. Now, no movement log is created for locations with **no beds**. Additionally, if an **empty `bedNames` array** is passed but the `plantAsset` does have beds associated with it, **all beds are chosen by default for termination**. Locations with active plant assets but no beds default to **100% area**, and **partially chosen beds do not contribute to the area percentage**.

This PR is ready for review.

modules/farm_fd2/src/entrypoints/soil_disturbance/lib.submitError.unit.cy.js
braughtg: For the quantities, this was because the log that references the quantities was deleted and farmOS deletes the quantities automatically in that situation.

braughtg:
The above all seem reasonable to me.

> With the latest update, if a plant asset has no beds, a movement log is no longer created. Previously, if the movement log failed, the archival would also not take place because it came after the log creation step. However, now that the movement log **never runs** for plant assets without beds, we can’t rely on its failure to prevent the plant from being archived when an error occurs.

This makes good sense.  A movement log would have failed due to a network error, if there is such an error then the archive will fail for the same reason.  The `createSoilDisturbanceTerminationLog` function is called from inside an operation in the entrypoint's `lib.js`.  So if the archive fails, it will then be handled by the `undo` part of the operation.

braughtg:
The code all looks good, but I am seeing two test failures when I run:
```
test.bash --unit --fd2 --glob=modules/farm_fd2/src/entrypoints/soil_disturbance/lib.submit.unit.cy.js 
```

```
1) Test the Soil Disturbance lib submission with activePlantAssets=true termination=true
       Checks the termination activity--log 1:
     TypeError: Cannot read properties of undefined (reading 'type')
```

```
2) Test the Soil Disturbance lib submission with activePlantAssets=true termination=true
       "after all" hook for "Check the soil disturbance log--activity 1 1":
     TypeError: Cannot read properties of undefined (reading 'id')

Because this error occurred during a `after all` hook we are skipping the remaining tests in the current suite: `Test the Soil Disturbance l...`
```

I suspect this is due to not creating that movement log, but should be investigated and cleaned up.

Shahir-47:
@braughtg

Fixed the issue in soil_disturbance/lib.submit.unit.cy.js. I had set the movement log’s date to a future date, which prevented it from being registered today and caused the related issue. It should work correctly now.

braughtg (approved):

braughtg:
Pushed a few small changes:
- Simplified the comment for multiple passes with multiple assets by removing the counter for the asset.
- Added note to the movement (termination) log indicating which beds were terminated.

PR #393, opened 2024-10-16
NumericInput Component: Increment on empty field populates the field

### LINKED ISSUE #326: NumericInput Component: Increment on empty field bug
## Observed Behavior

If a `NumericInput` component has an empty value (`''`) the first click on any of the increment buttons sets its value to the minimum value for the component. 

- Go to Tray Seeding
- Click in and clear the "Trays" field.
- Click on any of the increment buttons.
- Observe the value as 0.01.
 
## Desired Behavior

If the field is blank its value should become the value of the increment if it is valid, and remain blank if it is not.  For example, clicking on a large increment with value 10, should set the field's value to `10` not to the minimum value. 

If the field is blank and a decrement button is clicked the behavior is less clear:
- Should it set the value to the minimum?
- Should it leave the value blank?
- This should be discussed here before implementation.

The behavior when the field is not blank should not be changed.

RolandLocke:
**Pull Request Description**

Adjusts first click behavior so that value is rounded to nearest increment.
When the current field is cleared the current field, then the increment amount they click populates the field.
Refactors NumericInput component for cleaner implementation.

Closes #326 

---

**Licensing Certification**


RolandLocke:
Okay there is something weird going on here in the testing of this component. Current test is written for the component, but has been commented so that the progress on this feature can be saved. 

Here is the issue:
Expected behavior is seen on the browser.

however, in Cypress this behavior is seen:
1. User clicks input field
2. User clears the field
3. Input is Blurred
4. User clicks the increment button, the value in the field jumps to Nan
5. If the increment button is clicked again, the value jumps to the correct value.

This seems related to the issue discussed in the [this pull request](https://github.com/FarmData2/FarmData2/pull/386) but I was unable to track down the root issue.

components/NumericInput/NumericInput.vue
```diff
@@ -278,13 +278,22 @@ export default {
       return this.valueAsString == null || this.valueAsString.length == 0;
     },
     disableSmallDec() {
-      return this.numericValue - this.incDecValues[0] < this.minValue;
+      return (
+        this.numericValue - this.incDecValues[0] < this.minValue ||
+        isNaN(this.numericValue)
```
braughtg: I am not sure that disabling the decrements when the field is blank doesn't necessarily make sense, particularly if the value can be negative.

It is probably worth thinking this through a little bit.  When does it make sense for the increment or decrement buttons to be disabled?

braughtg:
> This seems related to the issue discussed in the [this pull request](https://github.com/FarmData2/FarmData2/pull/386) but I was unable to track down the root issue.

I think you are right that this is related to the fact that the `value` `prop` is not `v-bind`ed in the component test, where it is in the examples page and in all of the endpoints (e.g. TraySeeding, etc).

Looks to me like you need to add a special case to the `adjustValue` function for when the field `isEmtpy` and `isValid`.  This can happen when the field is not required and has been edited to be empty.  Currently in that case the code tries to `parseFloat(this.valueAsString)` which is empty (`""`) and results in the `NaN` that you are seeing.

components/NumericInput/NumericInput.vue
RolandLocke: Now, assuming that a negative value might be entered. Then, if `Value` is currently blank (Nan) and `minValue` is a negative number then the decrement buttons that don't subceed `minValue` will remain enabled. Otherwise, the decrement buttons remain disabled. This covers cases where the user wants to enter a negative number and the `minValue `allows it and the more traditional case where the user wants to clear the field before entering a value.

RolandLocke:
> Looks to me like you need to add a special case to the `adjustValue` function for when the field `isEmtpy` and `isValid`. This can happen when the field is not required and has been edited to be empty. Currently in that case the code tries to `parseFloat(this.valueAsString)` which is empty (`""`) and results in the `NaN` that you are seeing.

Yes, this was the case. Changes were made to the `adjustValue` function.

braughtg (changes requested):
Everything here looks good.  Well done!  

I think there is one other case that we didn't catch.  If the user deletes the field contents, then all of the increment/decrement buttons become enabled even if the increment that they adjust by would take the value over or under the max or min.

To see this on the example page, set the minValue to -10 and the maxValue to 10. Then delete the contents of the field.  All of the increment and decrement buttons will be enabled, even though the large increments would try to adjust by 100.  If you click them they set the value to 10, which is reasonable. But it would probably be better to have the buttons disabled if they will would try to take the value out of range.

braughtg (changes requested):
The decrement buttons are now being disabled when a decrement will take the value below the minimum value.  Nicely done. 

I am however observing one strange behavior on the `NumericInputExample` page.  The following sequence of steps will illustrate it:
1. Refresh the page
2. Decrease the `minValue` by 10 to make it -10.
3. Click the medium decrement button to make the field value -10.
4. Decrease the `maxValue` by 100 to make it 900.
5. Click the medium increment button.  This should make the field value 0, but it makes it 10.

I cannot reproduce this behavior on other pages because we do not have the ability to change the max/min values interactively on those pages.  So I'm not sure if this is something unique to the Example page or if there is something in the component that we can touch up to correct that behavior.

RolandLocke:
@braughtg 
The behavior mentioned above is due to the component property updates arriving in the component from the example page.

In the `numericInput` component the `initialValueChanged` is stored under `data()`.

In the `numericInput` example page, when the user updates `maxValue` the component properties are updated:

` v-on:click="
              () => { `\
                `        this.maxValue += 100;` \
                `        this.componentKey++;` \
             ` }` \

Any data for the `numericInput` component not stored as a property in the example page will be reverted to its original component state. For `initialValueChanged`, that means it will be set to false, as it is listed under `data()`, each time the component's properties are updated.

To solve this on the example page, I added a `initialValueChanged` property. Anytime a property is updated on the example page where continuity of `initialValueChanged` should be preserved, it is set to false in the component:

` v-on:click=" `\
              `() => {` \
                `         this.maxValue += 100;` \
                `         this.initialValueChanged = true;` \
               `          this.componentKey++;` \
              `}` \

Granted this is not a perfect solution, but without further guidance I'm not sure how to preserve this continuity across the example page and the numeric input.

In this commit, I also allowed large decrements to overwrite the current value. Example:

1. Initial value = 1
2. Decrement value by 10
3. Value becomes -10

or

1. Initial value = 1
2. increment value by 1
3. Value becomes 2
4. Decrement value by 10
5. Value becomes -8

braughtg:
@RolandLocke I spent some time working on this today.  I think that everything is working now (except the tests).  I came up with a different solution to dealing with the data resetting when a `prop` is changed problem.  It uses the `$options` element in Vue to create some non-reactive data.  I also touched up the adjust value function so that it will always round to the nearest increment if the initial value has not been changed.  So this is slightly different than we had discussed.  For example:
- An initial value of 3 when incremented by 10 will become 10 (because 13 is closer to 10 than 20).
- An initial value of 6 when incremented by 10 will become 20 (because 16 is closer to 20 than 10).
- An initial value of 3 when decremented by 10 will become -10.
- An initial value of 6 when decremented by 10 will become 0.

I pushed these changes to this branch, but I didn't run the tests.  I'm sure that this broke some things.  So those will need to be fixed up.  Let me know if you want to take those on.  ~If not I will probably spend some time on that tomorrow.~

RolandLocke:
Changes made: 

1. Moved `this.$options.valueChanged = true` from the `watch: value()` function to the `watch: valueAsString()` function as cypress was not seeing the changes to the `value()` function.
2. `FormInput` was calling the incorrect function in the `on:update:model-value='valueChanged' causing a browser Type Error and test fails.
3. Edited a test to specify number of decimal place in the input field

Note: due to the new rounding method, if the initial value is .5 and the user clicks the +1 increment, the value will jump to 2. I can see the benefits(consistency) and the negatives (not immediately clear behavior) of this approach. We might want to consider ignoring the rounding cases when the increment is smaller than 1.

braughtg:
> We might want to consider ignoring the rounding cases when the increment is smaller than 1.

Good point.  Though, we don't currently have any use cases where the increment is smaller than 1.  So maybe best to let this be for now and see if the case ever arises.

braughtg:
I made a few small updates to the component to address another odd case I'd seen where the `update` event was not emitted when the min/max values were changed.  So I think everything with the component works now.  I also edited the descriptions for a few of the test cases to be a little more descriptive.

There are just a few test updates that I think we should make and it will be ready to merge.
- The tests for the rounding behavior should be clarified.  
  - I suggest we remove the following three tests:
    - Input of value larger than initial value replaces initial value
    - Input of value smaller than initial value adds to initial value
    - Input of value smaller than initial decimal value adds to initial value
  - Then replace them with tests that check that initial clicks on increment and decrement buttons will round appropriately given the initial value:
    - Check that increment rounds up (e.g. 6 -> 20)
    - Check that increment rounds down (e.g. 3 -> 10)
    - Check that decrement rounds up (e.g. 6 -> 0)
    - Check that decrement rounds down (e.g. 3 -> -10)
    - These can be checked just using the medium increment button since the code is the same for all buttons.

Tests should be added to `events` to check that `update:value` is emitted when:
  - A change to the `minValue` prop causes the value to change (e.g. `value` is 10 and `minValue` is changed to 20).
  - A change to the `maxValue` prop causes the value to change.

RolandLocke:
Removed the following tests: 
	1. Input of value larger then initial value replaces the initial value
	2. Input of value smaller than initial value adds to t initial value
	3. Input of value smaller than initial decimal value adds to initial value
Added the following tests:
	1. Increment button rounds up appropriately
	2. Increment button rounds down appropriately
	3. Decrement button rounds up appropriately
	4. Decrement button rounds down appropriately
Note: that this the rounding occurring for each button takes place with the same logic.

New Test `Emits "update:value" when value is affected minValue change` should fail and is related to reactivity issues seen across this whole component.

RolandLocke:
@braughtg solved the issue last night this is how I got the local copy to work in the test and the example page:
> minValue(newValue) {
      this.localMinValue = newValue;
      this.valueAsString = this.formatter(this.value);
      this.numericValue = this.value;
    },


I didn't push changes because I think this needs to be totally re-worked.

I think part of what is making working on this component so confusing is storing the current value in two places: both `numericValue` and `valueAsString`. I think that a better solution would be either:

1. Storing the current value as a `Number` in `numericValue` then `watch()`ing it assigning it's formatted changes to the `valueAsString`
2. Updating `valueAsString` wherever possible the downside being: the variable has to be parsed whenever needed for an operation

braughtg:
> this is how I got the local copy to work in the test and the example page:
> 
> > minValue(newValue) {
> > this.localMinValue = newValue;
> > this.valueAsString = this.formatter(this.value);
> > this.numericValue = this.value;
> > },

@RolandLocke It makes sense to me that that works given the way the component is structured now.  

I also agree that thinking about the internal organization of this component makes sense to try to simplify.  It feels like we are working against or around Vue here.  Your idea about how to store the value only once makes sense, but I feel like I did that in the first place to help deal with this issue.  So I think once we get a better handle on why the component is not reactive in the way we want it to be we'll land on a better solution.  I plan to do some reading and research on this today and will report back.

braughtg:
@RolandLocke I've made some progress on this but I am running into another odd behavior where some watcher's don't seem to be getting called (e.g. if I watch `minValue` or `maxValue`) the watchers are not invoked when those props are changed.  But if I watch `required` its watcher is invoked any time that prop is changed.  I am not sure what's going on with that.  I will continue to probe this issue over the next few days.

braughtg:
@RolandLocke  Okay... I think what I pushed to this branch should be good.  I refactored it to use just the numeric value as state as you had suggested. I also made some changes to ensure that that value is always bound to [`minValue`, `maxValue`] or is `NaN`, which simplified the logic elsewhere quite a bit.  

Turns out that a lot of the issues with reactivity that we were seeing were due to the example page using the component `key` to try to force updates.  That was somehow preventing the watchers on some props from running.  I don't fully know why, but when I removed those parts from the example page, all the watchers started being called just fine. 

The code currently passes all of the tests that are there.  I removed a bunch of styling tests because they no longer make sense since the value of the component can no longer be anything but a valid value or `NaN`. I also touched up a few other tests here and there.  

I didn't review the tests that you had added for this issue and for the rounding behavior yet.  You should review and ensure that those are completed.  Then also, have a look over the new implementation and see if it all makes sense to you.  

Once everything looks good to you, flip it to a full PR and I'll review it.

braughtg (approved):

RolandLocke:
There is an only on the events test. Removing it reveals issues tests that are not passing.

braughtg:
@RolandLocke I now see the tests that are not passing.  Because of the earlier changes it is no longer possible for the input to contain an invalid value (less than `minValue`, more than `maxValue`, or a non-numeric string). As a result, I think instead of just trying to adapt the current tests we should rethink them.  Here is a proposed set of tests:
1. Initial `value` prop tests
1.a `value` < `minValue` - emits `update:value` with `minValue` and `valid` with `true`
1.b `value` > `maxValue` - emits `update:value` with `maxValue` and `valid` with `true`
1.c `value` is invalid (e.g. `abc`) - emits `update:value` with `NaN` and `valid` with `false`
1.d. `value` is valid (e.g. 7) - emits `update:value` with new `value` and `valid` with `true`
2. `value` prop change tests
2.a `value` changes from valid (e.g. 7) to valid (e.g. 8), emits `update:value` with new `value`, but does not emit `valid`
2.b `value` changes from valid to invalid (e.g. `abc`), emits `update:value` with `NaN` and `valid` with `false`
2.c `value` changes from invalid (e.g. `abc`) to invalid (`def`), does not emit `update:value` or `valid`
2.d `value` changes from invalid to valid, emits `update:value` with new `value` and `valid` with `true`
2.e `value` changes from valid to < `minValue`, emits `update:value` with `minValue`, but does not emit `valid`
2.f `value` changes from valid to > `maxValue`, emits `update:value` with `maxValue`, but does not emit `valid`
2g `value` changes from invalid to < `minValue`, emits `update:value` with `minValue` and `valid` with `true`
2h `value` changes from invalid to > `maxValue`, emits `update:value` with `maxValue` and `valid` with `false`

I think that gives us pretty good coverage.  Let me know if you can think of any others or if any of these don't make sense.  If all looks good, go ahead and revise `NumericInput.behavior.comp.cy.js` to have these tests.

Also, be sure to pull your branch as I pushed a few small changes.

RolandLocke:
@braughtg are we changing the `value` prop its self in the 2.x tests or are we updating the `numeric-input` field? both cases need to emit the valid and `update:value` events. Maybe a third set of tests repeating the 2.x tests that check `value` prop updates or  `numeric-input` field updates? Or maybe the field `valid`/`update:value` tests should be added to the behavior tests? Please let me know.

I've noticed the following tests were missing:

1.e `required` is false, `value` is invalid (e.g. abc) - emits `update:value` with NaN and `valid` with true
(this one might not be necessary) 

2.i `maxValue` changes (100 -> 10) to exclude `value` (50), emits `update:value` with `maxValue` and `valid` with true
2.j `minValue` changes (-50 -> 0) to exclude `value` (-10), emits `update:value` with `minValue` and `valid` with true

RolandLocke:
I added the Following tests as suggested:

1. Initial value prop tests
  1.a value < minValue - emits update:value with minValue and valid with true
  1.b value > maxValue - emits update:value with maxValue and valid with true
  1.c value is invalid (e.g. abc) - emits update:value with NaN and valid with false
  1.d. value is valid (e.g. 7) - emits update:value with new value and valid with true
  (additional see[ comment](https://github.com/FarmData2/FarmData2/pull/393#issuecomment-2501275229)) 1.e required is false, value is invalid (e.g. abc) - emits update:value with NaN and valid with true
2. value prop change tests
  2.a value changes from valid (e.g. 7) to valid (e.g. 8), emits update:value with new value, but does not emit valid
  2.b value changes from valid to invalid (e.g. abc), emits update:value with NaN and valid with false
  2.c value changes from invalid (e.g. abc) to invalid (def), does not emit update:value or valid
  2.d value changes from invalid to valid, emits update:value with new value and valid with true
  2.e value changes from valid to < minValue, emits update:value with minValue, but does not emit valid
  2.f value changes from valid to > maxValue, emits update:value with maxValue, but does not emit valid
  2g value changes from invalid to < minValue, emits update:value with minValue and valid with true
  2h value changes from invalid to > maxValue, emits update:value with maxValue and valid with false
   (additional see[ comment](https://github.com/FarmData2/FarmData2/pull/393#issuecomment-2501275229)) 2.i maxValue changes (100 -> 10) to exclude value (50), emits update:value with maxValue and valid with true
  (additional see[ comment](https://github.com/FarmData2/FarmData2/pull/393#issuecomment-2501275229)) 2.j minValue changes (-50 -> 0) to exclude value (-10), emits update:value with minValue and valid with true

In `NumericInput` I updated some linting issues that I encountered.

braughtg:
@RolandLocke 

> are we changing the `value` prop its self in the 2.x tests or are we updating the `numeric-input` field? 

We will only need these tests to check that the correct events are emitted when the `value` prop is changed.  We don't need to test typing in the field because when text is typed in the field the `set` method in the `formattedValue` computed property is used to update the `numericValue`. That `set` method uses the `validateValue` method to ensure that the entered value is validated in the same way that it is when the `value` prop is changed (see the watcher for `value`).

> I've noticed the following tests were missing:
> 
> 1.e `required` is false, `value` is invalid (e.g. abc) - emits `update:value` with NaN and `valid` with true (this one might not be necessary)

Good catch. Adding this one makes sense. 

> 
> 2.i `maxValue` changes (100 -> 10) to exclude `value` (50), emits `update:value` with `maxValue` and `valid` with true 
> 2.j `minValue` changes (-50 -> 0) to exclude `value` (-10), emits `update:value` with `minValue` and `valid` with true

Also worth having these.  Nice work.

PR #397, opened 2024-10-29
fix(comp): prevents duplicate equipment selection in EquipmentSelector

### LINKED ISSUE #164: EquipmentSelector: Disallow duplicate equipment
The `EquipmentSelector` currently allows the same piece of equipment to be selected multiple times.  

For example, in the Direct Seeding form:

image

This issue should be addressed by modifying the `EquipmentSelector` so that the same piece of equipment cannot be selected multiple times.  This could be done by disabling items that have already been selected or not having them appear at all in subsequent `Selector` elements.  Consideration should be given to which of those two approaches is best with discussion occurring in this issue's comments.

~~Note: Currently, if the same piece of equipment is selected multiple times, then if the first one is deleted the last one will also disappear and the equipment selector will be left in an invalid state with multiple blank `Selector` elements.  For example:~~

image

~~Disallowing multiple selections of the same equipment will fix this issues as a matter of course.~~

Shahir-47:
### Pull Request Description

This pull request fixes issue #164 by preventing duplicate selections in the `EquipmentSelector` component. Rather than directly modifying `EquipmentSelector`, I updated the `MultiSelectorBase` component, which manages the dropdown behavior for multiple selections within `EquipmentSelector`.

#### Changes Made
1. **Added a `noRepeat` Prop**:
   - I added a `noRepeat` prop to `MultiSelectorBase`, set to `false` by default so it doesn’t interfere with other components using `MultiSelectorBase`. When `noRepeat` is set to `true`, the dropdowns automatically filter out any items that have already been selected, ensuring each item can only be chosen once.
   - In `EquipmentSelector`, I enabled this feature by setting `v-bind:noRepeat="true"`, with no further changes needed in `EquipmentSelector`.

2. **Updated the Dropdown Filtering Logic**:
   - I created a new method, `getFilteredOptions`, in `MultiSelectorBase` to handle filtering when `noRepeat` is enabled. This method updates each dropdown to hide items already selected, allowing only unique selections across dropdowns.

3. **Added Behavior Tests in `MultiSelectorBase.behaviour.comp.cy.js`**:
   - I added new behavior tests to `MultiSelectorBase.behaviour.comp.cy.js` to verify that `noRepeat` works as expected. These tests check that selected items are removed from the available options in subsequent dropdowns and reappear when deselected, confirming that the `noRepeat` functionality works smoothly.

#### Why I Updated `MultiSelectorBase`
The issue #164 suggested modifying `EquipmentSelector`, but since `MultiSelectorBase` controls the dropdowns, adding `noRepeat` there made the solution simpler and more flexible.

Closes #164 

---

**Licensing Certification**

### DIFF
--- a/components/EquipmentSelector/EquipmentSelector.vue
+++ b/components/EquipmentSelector/EquipmentSelector.vue
@@ -8,6 +8,7 @@
       v-bind:showValidityStyling="showValidityStyling"
       v-bind:options="equipmentList"
       v-bind:selected="selected"
+      v-bind:noRepeat="true"
       v-on:update:selected="handleUpdateSelected()"
       v-on:add-clicked="handleAddClicked($event)"
       v-on:valid="handleValid($event)"
--- a/components/MultiSelectorBase/MultiSelectorBase.behavior.comp.cy.js
+++ b/components/MultiSelectorBase/MultiSelectorBase.behavior.comp.cy.js
@@ -192,4 +192,239 @@ describe('Test the MultiSelectorBase component behavior', () => {
         cy.get('[data-cy="selector-4"]').should('not.exist');
       });
   });
+
+  it('Selecting an option removes it from available options in subsequent selectors', () => {
+    const readySpy = cy.spy().as('readySpy');
+
+    cy.mount(MultiSelectorBase, {
+      props: {
+        onReady: readySpy,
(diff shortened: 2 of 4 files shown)


braughtg:
See my comment on #164.

PR #402, opened 2024-11-20
feat(dev): added context sensitive button for PickerBase and PicklistBase

### LINKED ISSUE #166: PickerBase: Add a context sensitive icon to "All" button
__Current Behavior__

Currently the `PickerBase` component displays an "All" button. This button can be seen in the `BedPicker` on the Direct Seeding form (which is created using a `PickerBase` component)  by selecting a Location that has beds (e.g. `CHUAU`).

image

Clicking the "All" button either selects all of the beds (if they are not all selected) or deselects all beds (if they are all selected).  

__Desired Behavior__

The `PickerBase` class should have a button that displays something context sensitive that indicates what will happen if it is clicked.  Maybe we could use an icon like:  
:white_check_mark: All (when it will select all)
or
:no_entry_sign: All (when it will clear all)

Other proposals for the design of the buttons should be proposed here for discussion before implementation.

### LINKED ISSUE #186: PicklistBase: Add a context sensitive icon to "All" button
See #166 which describes the issue for `PickerBase`.  The same issue applies to the `PicklistBase` component.

Image

The solution used should be be consistent with each other.

WarpWing:
**Pull Request Description**

Adds context sensitive button for PickerBase and Picklist Base: See #166 and #186 

Also added some styling on top of that logic to adjust for icon shift.

Closes #166
Closes #186 

---

**Licensing Certification**


components/PickerBase/PickerBase.vue
```diff
@@ -23,14 +23,17 @@
...
+           id="picker-all-button"
+           data-cy="picker-all-button"
+           size="sm"
+           :variant="checked.length === options.length ? 'danger' : 'primary'"
+           v-on:click="pickAll"
+          >
+           <span v-if="checked.length === options.length">🚫 All</span>
+           <span v-else>✅ All</span>
+         </BButton>
           >
```
braughtg: There is an extra `>` here that is showing up in the UI and it should be removed.

components/PickerBase/PickerBase.vue
```diff
@@ -23,14 +23,17 @@
...
-            size="sm"
-            variant="primary"
-            v-on:click="pickAll()"
-            >All</BButton
+         <BButton
+           v-if="showAllButton"
+           id="picker-all-button"
+           data-cy="picker-all-button"
+           size="sm"
+           :variant="checked.length === options.length ? 'danger' : 'primary'"
```
braughtg: The `danger` and `primary` variants seem a bit overly intrusive to me.  I would just leave this as `primary` and let the icon be sufficient for indicating what will happen. The PickListBase doesn't use the red and green variants and I prefer it.

components/PicklistBase/PicklistBase.vue
```diff
@@ -25,7 +25,8 @@
                 variant="primary"
                 v-on:click="handleAllButton()"
               >
-                All
+                <span v-if="allPicked">🚫 Clear All</span>
+                <span v-else>✅ Select All</span>
```
braughtg: Leave the text here as just "All".  It is a little less clear, but also takes up less screen real estate. We are designing to run in portrait mode on small mobile (iPhone SE in devtools). Using less text here allows more room for the other columns.

components/PicklistBase/PicklistBase.vue
```diff
@@ -36,7 +37,8 @@
                 variant="primary"
                 v-on:click="handleUnitsButton()"
               >
-                {{ units }}
+                <span v-if="allPicked">🚫 Clear {{ units }}</span>
```
braughtg: Leave the text here as just `{{ units }}` for the same reason as expressed above.

braughtg (changes requested):
Overall this is looking pretty good.  I do have a few thoughts.  See my comments in the code.

The testing for this should really only require that you modify the existing tests to check for the presence of the appropriate icons at the appropriate points.

WarpWing:
@braughtg I've made the appropriate changes. I'll work on modifying existing testing and will then change PR state from draft. Let me know if you need anything further.

components/PickerBase/PickerBase.content.comp.cy.js
```diff
@@ -174,4 +174,97 @@ describe('Test the default PickerBase content', () => {
...
+      
+      cy.get('[data-cy="picker-all-button"]').should('contain.text', '✅ All');
+
+      cy.get('[data-cy="picker-all-button"]').click();
+
+      cy.get('[data-cy="picker-options"]').find('input').should('be.checked');
+      
+      cy.get('[data-cy="picker-all-button"]').should('contain.text', '🚫 All');
+
+      cy.get('@updateSpy').should('have.been.calledWith', [
```
braughtg: This assertion is not necessary here as it is tested in the `events` tests.  Similar comment applies to other tests here that check the `picked` event payload.

braughtg (changes requested):
This is starting to look good.  I left one comment in the code and have a few more here.

- [ ] - The button in the `PickerBase` is showing up with `variant` `second` (grey) instead of `variant` `primary` (blue).  The variant in the `PicklistBase` is correct (blue). We can debate which variant is the right choice.  But they should be consistent.  I didn't see an obvious reason for this as the code looks correct, but the button is still showing up incorrectly.  This will need some diagnosis.
- [ ] - The tests that you added belong in the `.behavior.comp.cy.js` file since they are testing the behavior of the component rather than its initial content. 
- [ ] - On the example page for the `PickListBase` component there is a button for "Pick row" that doesn't seem to be working correctly. This button should change the `picked` prop so the first row is either picked or not (toggle it). Currently when no rows are selected, clicking this button doesn't seem to have any effect.  However, when the first row is selected, clicking the button will deselect the row.  This could be a problem with the example page or with the component. If it is with the example page we can just fix it.  If it is in the component then we should try to add a test to the `.behavior` file that fails in this condition, and then try to fix it.

DuckySoLucky (approved):
lgtm, 30, push to prod

> ignore this for legal reasons

WarpWing:
@braughtg I've written PickerBase and PickListBase behavior and content tests as well as fixed the primary vs variant issue as well. I will look at the picked problem and open a separate [PR](https://github.com/FarmData2/FarmData2/pull/408)

components/PickerBase/PickerBase.content.comp.cy.js
```diff
@@ -174,4 +174,73 @@ describe('Test the default PickerBase content', () => {
...
+
+    cy.get('[data-cy="picker-all-button"]')
+      .should('contain', '✅ All')
+      .click()
+      .should('contain', '🚫 All')
+      .click()
+      .should('contain', '✅ All');
+  });
+
+  it('Test All button text updates when options are selected individually', () => {
```
braughtg: This test seems redundant with the similar test in `behavior`.  It is testing the _behavior_ of the component so we should keep the one in `behavior` and delete this one.

components/PickerBase/PickerBase.content.comp.cy.js
```diff
@@ -174,4 +174,73 @@ describe('Test the default PickerBase content', () => {
...
+        options: ['Option 1', 'Option 2', 'Option 3', 'Option 4'],
+        picked: ['Option 1', 'Option 2', 'Option 3', 'Option 4'],
+        invalidFeedbackText: 'Invalid feedback text.',
+      },
+    });
+
+    cy.get('[data-cy="picker-all-button"]').should('contain', '🚫 All');
+  });
+
+  it('Test All button text toggles when clicked', () => {
```
braughtg: This test seems redundant with the similar test in `behavior`. It is testing the _behavior_ of the component so we should keep the one in `behavior` and delete this one.

components/PickerBase/PickerBase.behavior.comp.cy.js
```diff
@@ -226,4 +226,111 @@ describe('Test the PickerBase component behavior', () => {
         });
     });
   });
-});
+
+  it('should initially show "✅ All" when no items are selected', () => {
```
braughtg: This test is redundant with the similar one in `content`.  It is checking the _content_ of the component when it is rendered.  So it belongs in `content` so this one should be deleted and the one in `content` should be kept.

components/PickerBase/PickerBase.behavior.comp.cy.js
```diff
@@ -226,4 +226,111 @@ describe('Test the PickerBase component behavior', () => {
...
+        showAllButton: true,
+        picked: []
+      }
+    });
+
+    cy.get('[data-cy="picker-all-button"]')
+      .should('contain', '✅ All');
+  });
+
+  it('should show "🚫 All" when all items are selected', () => {
```
braughtg: This test is redundant with the similar one in `content`.  It is checking the _content_ of the component when it is rendered.  So it belongs in `content` so this one should be deleted and the one in `content` should be kept.

components/BedPicker/BedPicker.vue
```diff
@@ -10,6 +10,8 @@
       v-bind:picked="picked"
       v-bind:required="required"
       v-bind:showValidityStyling="showValidityStyling"
+      v-bind:allButtonVisible="true"
```
braughtg: This `prop` does not exist in `PickerBase` and should be removed.

components/BedPicker/BedPicker.vue
```diff
@@ -10,6 +10,8 @@
...
+      v-bind:allPicked="isAllPicked"
```
braughtg: This `prop` does not exist in `PickerBase` and should be removed.

components/BedPicker/BedPicker.vue
```diff
@@ -95,7 +97,11 @@ export default {
       bedList: [],
     };
   },
-  computed: {},
+  computed: {
+    isAllPicked() {
```
braughtg: I don't think we need this `computed` property.  The `prop` that it was bound to in `PickerBase` doesn't exist so I don't think this has any effect.  Please respond to this comment if I am missing something.  Otherwise, remove this `computed` property.

components/PicklistBase/PicklistBase.vue
```diff
@@ -36,7 +36,8 @@
                 variant="primary"
                 v-on:click="handleUnitsButton()"
               >
-                {{ units }}
+                <span v-if="allPicked">🚫 {{ units }}</span>
```
braughtg, line 40: We need something similar above for the "All" button when `units` are not being used.  To see this go to the `PicklistBase` example page and click the "useUnits" `prop` control to switch to not using `units`. In that case the button just contains "All".  It should have the icon same as when we are using `units`.

The test cases will need to be updated to also include tests for the context sensitive icon when `units` are not being used.  They will parallel the ones for when `units` are being used.

components/PicklistBase/PicklistBase.content.comp.cy.js
```diff
@@ -11,283 +11,156 @@ describe('Test the default PicklistBase content', () => {
...
-          .should('exist')
-          .and('contain.text', 'Name');
-        cy.get('[data-cy="picklist-sort-button-quantity"]')
-          .should('exist')
-          .and('contain.text', 'Quantity');
-        cy.get('[data-cy="picklist-sort-button-location"]')
-          .should('exist')
-          .and('contain.text', 'Location');
-      });
+    cy.get('[data-cy="picklist-all-button"]')
```
braughtg: This looks like the right test. But it won't pass because the icon doesn't exist in the button when in _checkbox_ mode (i.e. when not using `units`).  See the comment in `PicklistBase.vue`.

components/PicklistBase/PicklistBase.content.comp.cy.js
braughtg, line 1: This file should be reverted to its original content and then only small changes should be made to test just the presence of the context sensitive icon when `units` is being used and when `units` is not being used. As you did with `PickerBase` the tests should check for ✅ when no rows are picked and 🚫 when all rows are picked (using the `picked` `prop`).

components/PicklistBase/PicklistBase.behavior.comp.cy.js
```diff
@@ -142,4 +142,82 @@ describe('Test the PicklistBase component behavior', () => {
...
+    });
+
+    cy.get('[data-cy="picklist-units-button"]').contains('✅ Trays');
+    cy.get('[data-cy="picklist-units-button"]').click();
+    cy.get('[data-cy="picklist-units-button"]').contains('🚫 Trays');
+    cy.get('[data-cy="picklist-units-button"]').click();
+    cy.get('[data-cy="picklist-units-button"]').contains('✅ Trays');
+  });
+
+  it('hides header button when showAllButton is false', () => {
```
braughtg: This is checking the _content_ of the rendered component based on a `prop`, so this test should be moved to the `content` tests and deleted here.  This should also be tested both when `units` are being used and when they are not.

components/PicklistBase/PicklistBase.behavior.comp.cy.js
```diff
@@ -142,4 +142,82 @@ describe('Test the PicklistBase component behavior', () => {
...
+        showInfoIcons: true
+      },
+    });
+
+    cy.get('[data-cy="picklist-info-icon-0"]').click();
+    cy.get('[data-cy="picklist-all-button"]').should('be.disabled');
+    cy.get('[data-cy="picklist-checkbox-0"]').should('be.disabled');
+  });
+
+  it('toggles between checkmark and x icons for All button', () => {
```
braughtg: This looks like a good test but it will not pass because the icon is not implemented when `units` are not being used.  See comment in `PicklistBase.vue`.

braughtg (changes requested):
I've left a number of comments throughout the code.  Please respond to any that are unclear.  Also, be sure to run the tests when you think you have things working and made sure that they all pass.

To run the tests on this component use:
`test.bash --comp --glob=components/PickerBase/*.comp.cy.js` (or something similar for the other components.)

WarpWing:
Checkpoint: I've gotten through behavior, content and event tests for `PickerBase`.  Taking a short 10 minute break before proceeding with `PicklistBase` tests.
image

WarpWing:
Checkpoint: added behavior tests to Picklistbase, not sure if I need any content or event tests, I'll look at it more tonight or tomorrow morning.
image

WarpWing:
@braughtg flipped the important one to ready for review, revamped the tests, if you would like any specific tests for me to make, let me know.

PR #403, opened 2024-11-21
fix(comp): adds support to disable object in selector base.

### LINKED ISSUE #399: SelectorBase Component: Add support for 'options' containing a list of objects with 'disabled' property
The `SelectorBase` component currently accepts a list `options` which contains strings. Each string becomes a selectable item in the drop-down. 

Changes in [issue #164](https://github.com/orgs/FarmData2/projects/2?pane=issue&itemId=78337354&issue=FarmData2%7CFarmData2%7C164) will eventually provide the `SelectorBase` component with a list of objects with the following structure:

`{ text: "...", value: "...", disabled: "true/false" }`

This format is based on the Bootstrap-Vue [Form Select Component's `options` format](https://bootstrap-vue.org/docs/components/form-select). `SelectorBase` should both support `options` as a list of strings and as a list of objects. 

If the `SelectorBase` component receives `options` as a list of strings, it should:

1. Convert the array of strings given as `options` to a lists of objects with the above format to populate the list.
2. Emit the `selectedOption` as a string, as it does currently.

If the `SelectorBase` component receives `options` as a list of objects using the above format, it should:

1. Populate the list using the `options` directly.
2. Emit the `selectedOption` as a string, as it does currently.

RolandLocke:
**Pull Request Description**

The `SelectorBase `component now can be provided either a list strings or a list o objects in the following structure:

{ text: "...", value: "...", disabled: "true/false" } 
(Format is based on the Bootstrap-Vue [Form Select Component's options format](https://bootstrap-vue.org/docs/components/form-select))

If the `SelectorBase `component receives options as a list of strings:

Converts the array of strings to a lists of objects with the above format to populate the `optionslist`.
Emits the `selectedOption `as a string.

If the `SelectorBase` component receives options as a list of objects using the above format:

Populate the `optionslist `using the options directly.
Emit the `selectedOption `as a string.

To achieve this, `options` lists passed to the component through the `options` prop are normalized into the `text`:, `value`:, `disabled`: format. If the object is already in the expected format, it will not update the list. If the object is not in the expected format, the program has no checks to ensure that it is. The `Selected` has not changed in functionality, it will return the `value`: of the `optionsList` selected object. Finally, if an object is passed to selector base with `disabled:` set to true, it will be greyed out and wont be selectable.

The following tests were added to test the functionality:

1. `Prop is a list of strings and objects` tests the ability of selector base to successfully create a options list from mixed types (objects and strings). It also tests the reactivity when updating the prop.
2. `Props reflect can be disabled/enabled through flag` tests that props are disabled and enabled as specified in the prop. It also tests that the props are updated properly when the disabled flags are updated.

Closes #399 

---

**Licensing Certification**


modules/farm_fd2_examples/src/entrypoints/selector_base/App.vue
braughtg, line 1: Is there a reason not to leave initial `selected` value as null?

The example page should be modified to allow an option to be disabled in the list.  Maybe add a button to the `options` row that will "Disable three" or "Enable three".

braughtg (changes requested):
Overall looks pretty good.  See my one comment on the example page.

Also, I think there should be a test in `content` just for the object based option list that is separate from the tests you have here that check that the prop change works.

modules/farm_fd2_examples/src/entrypoints/selector_base/App.vue
RolandLocke: No, there isn't a reason to leave initial `selected ` null I've changed it.

RolandLocke:
Added 'Check object props can be updated' test. Checking that an options list of objects can be updated via `disabled` properties only and reflect the changes.

Added support on the example page for disabling three objects from the example list. A new button Disable Three allows the user to toggle on and off the disabling of items. While creating this feature I had to ensure that if an object was selected, it would be unselected when the item was disabled. I did this in the example page button, but I'm wondering if this should be a feature in `selectorBase`. One possible solution:

- If the user selects an item that later becomes disabled
          - The selected item stays selected, and the component emits not valid + communicates to user that an item selectibility has changed.

By not "blanking" the field, the user can see what item they had previously selected and possibly revert the changes that caused it. I think this would also prevent data loss if the user had filled out the soil disturbance page and the option they had filled out was greyed. I also struggle to think of a good reason the user could disable an item later in the form, more thoughts on this might be helpful.

@braughtg let me know what you think of these changes and proposed change.

modules/farm_fd2_examples/src/entrypoints/selector_base/App.vue
```diff
@@ -112,6 +112,39 @@
...
+        <td>
+          <BButton
+            id="disable-button"
+            data-cy="disable-button"
+            variant="outline-primary"
+            size="sm"
+            v-on:click="
+              () => {
+                const itemsToDisable = ['one', 'two', 'four'];
+                if (itemsToDisable.includes(this.form.selected)) {
```
braughtg: This should definitely be part of the behavior of the component as you suggested.  We should decide exactly what that behavior should be and how it should be controlled in the context of a single `SelectorBase` and in the context of a parent component that uses multiple `SelectorBase` components (e.g. `MultiSelectorBase`).

modules/farm_fd2_examples/src/entrypoints/selector_base/App.vue
```diff
@@ -112,6 +112,39 @@
           </BButton>
         </td>
       </tr>
+      <tr>
+        <td>Disable Three Items</td>
```
braughtg: I wonder if this should be "Toggle Disabled Items" and the button will either disable or enable the items depending on their current state.

braughtg (changes requested):

RolandLocke:
In reference to this in my earlier comment:

> While creating this feature I had to ensure that if an object was selected, it would be unselected when the item was disabled. I did this in the example page button, but I'm wondering if this should be a feature in selectorBase

As discussed in a meeting, between @braughtg and I, a prop will be added to the `selectorbase` component to adjust the behavior of disabling a prop from the list. 

Behavior:
When a `selectorbase` component has an option selected, whose disabled attribute becomes `false`, 

If `keepDisabledSelected` prop is `true`:
The selected option remains selected.

If `keepDisabledSelected` prop is `false`:
The selected option is removed from the selected items.

When the `keepDisabledSelected` prop is `true` this will allow us to store a single `options` list in the `multiSelectorComponent` disabling options from `options` as the user selects them in the `selectorbase` components.

RolandLocke:
Added a prop to the component called `keepDisabledSelected` allowing the component to keep items selected that were disabled in the list.

Added appropriate tests checking that items are removed from the list when disabled and `keepDisabledSelected` is false. 

Added a toggle to the example page to test the functionality

Added a new method `checkSelectedOption()` that ensures that the `selected` item is exists in the list and is not disabled.

braughtg (changes requested):
This is looking pretty good.

I see one odd behavior in the example page:
- With "Keep Disabled Selected" false
- Click "Toggle Disabled Items"
- Click "Select First option"

The first option "one" will be selected, even though it is disabled.

Seems like this will require a check in the `watch` for `selected`.  I would suggest that you write a test for this case in `behavior` and see that it fails.  Then make the fix and ensure everything still passes.

RolandLocke:
Some extra items that need to be added to the change before completion:
1. Add docs string for the new prop.
2. Header for the docs needs example and the new prop.
3. optionsList and optionList need renaming (processedOptions?)

RolandLocke:
I added docs for the `keepDisabledSelected ` prop as well as an updated the example for the component HTML implementation in docs.

I made changes to the component data:

Removed `optionsList` as it was no longer necessary.

Changed name for clarity `optionList` -> `optionsObjects `

Added functionality to prevent selected prop cannot be changed to a disabled item and associated test. 
@braughtg please note that line 434 in SelectorBase seems strange to me and I'm not entirely sure why adding it fixed the issue.

components/SelectorBase/SelectorBase.vue
```diff
@@ -405,7 +426,15 @@ export default {
       this.$emit('valid', this.isValid);
     },
     selected() {
-      this.selectedOption = this.selected;
```
braughtg, line 408: Here's what I put in for `selected()` and it seems to be working:
```JavaScript
   selected() {
      for (let option of this.optionsObjects) {
        if (option.text === this.selected) {
          if (!option.disabled || this.keepDisabledSelected) {
            this.selectedOption = this.selected;
          } else {
            this.selectedOption = '';
            this.$emit('update:selected', this.selectedOption);
          }
          return;
        }
      }
    },
```

The reason that the `$emit` is needed here is because `this.selectedOption` was not changing if `this.selected` was disabled.  Thus, the `selectedOption` watcher was not invoked.  Above I've changed `this.SelectedOption` to be `''` which invokes the `selectedOption` watcher the first time the disabled option is selected. But if the disabled item is selected a second time then `selectedOption` would not change (it would already be `''`), so we need to emit here anyway.

braughtg: @RolandLocke See the comment above!

braughtg (approved):

braughtg:
@RolandLocke Looks like this has broken some of the test for other components that use the `SelectorBase` component.  I'm out of time to look at it today, but will come back to it tomorrow and try to diagnose the issue.  If you want to have a look before then you can run `test.bash --comp`.

braughtg:
I've been looking at this and haven't found an easy solution.  It seems that there is another reactivity issue when we place this component inside of another one.  I'll keep digging.

braughtg:
I think the last commit took care of it.  There was an issue with the initial `selected` prop was not getting rendered and a case where setting `selected` to the empty string was causing problems. Tests were added to the suite and the code was patched.  I'm seeing a little test flake in unrelated tests, so I'm going to let it sit until morning and run them again.  If all looks good I'll merge it.

PR #408, opened 2024-12-13
fix(examples): fixes mixed type sorting in PickListBase

### LINKED ISSUE #414: Improve PicklistBase sorting of mixed data types
Currently the sorting if data in the `PicklistBase` does not consider data type when comparing values in the column being sorted.  This results in some seemingly odd orderings when the columns contain mixed type data (e.g. strings and numbers).  The data types of the items being compared should be considered in the sort so that when:
- Two strings are compared they are compared lexicographically.
- Two numbers are compared they are compared numerically.
- When a string and a number are compared the number is converted to a string and they are compared lexicographically.

WarpWing:
**Pull Request Description**

Fixes the sorting of mixed type columns (e.g.  numbers and strings).  

Also fixes example page issue related to `picked` prop not being properly updated due to non trigger of reactivity and refactors `handleSort`.

Closes #414 

---

**Licensing Certification**

### DIFF
--- a/components/PicklistBase/PicklistBase.behavior.comp.cy.js
+++ b/components/PicklistBase/PicklistBase.behavior.comp.cy.js
@@ -61,7 +61,98 @@ describe('Test the PicklistBase component behavior', () => {
     cy.get('[data-cy="picklist-quantity-1"]').should('have.value', '0');
   });
 
-  it('sorts rows correctly and updates picked rows and quantities', () => {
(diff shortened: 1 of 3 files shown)


braughtg:
When reviewing this PR I noticed an issue with the sorting of the rows when the column contained mixed type data (e.g. a string and a number).  So I fixed that along with keeping the changes that were in the original PR.

PR #409, opened 2024-12-13
fix(dev): added persistent loop for authentication

### LINKED ISSUE #407: Troubleshoot script progression of `setup.bash` when presenting an invalid PAT during login
After running a fresh install on Gitpod, I was presented with the option to login via a PAT.
image

Though the PAT wasn't accepted (see error) message, the script marked me as authenticated and proceeded to build. In this case, I would imagine we should pause script progression until a valid way of authentication is accepted. 

I would recommend a fix like 
```bash
while ! gh auth login --hostname GitHub.com || ! gh api user &> /dev/null; do
    echo "Authentication failed. Clearing credentials and trying again..."
    gh auth logout --hostname GitHub.com &> /dev/null
    read -p "Press Enter to retry or Ctrl+C to exit..."
done
 ```
Waiting on maintainer thoughts before proceeding with a PR in case this conflicts with any design philosophies

WarpWing:
**Pull Request Description**

Closes #407 

---

**Licensing Certification**

### DIFF
--- a/bin/setup.bash
+++ b/bin/setup.bash
@@ -78,7 +78,11 @@ echo "      - How would you like to authenticate GitHub CLI?"
 echo "        - Login with a web browser"
 echo "          - Use your GitHub username and password to log in."
 echo ""
-gh auth login --hostname GitHub.com
+while ! gh auth login --hostname GitHub.com || ! gh api user &> /dev/null; do
+    echo "Authentication failed. Clearing credentials and trying again..."
+    gh auth logout --hostname GitHub.com &> /dev/null
+    read -p "Press Enter to retry or Ctrl+C to exit..."
+done
 echo "  Authenticated."
 
 echo ""
--- a/bin/setup.bash
+++ b/bin/setup.bash
@@ -67,21 +67,23 @@ echo "  Configured."
 echo ""
 
 echo "  Authenticating with GitHub..."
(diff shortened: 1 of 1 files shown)


braughtg (changes requested):
This has some good elements and ideas in it.  However, it doesn't seem to work correctly.  A few thoughts here:
- The `gh api user` command return 0 when logged in and non-zero when not logged in.  So I think the condition on that needs to not use `!`.  So the loop should be "while not logged in" or `while gh api user`.
- Then in the loop do the `gh auth login` inside the loop.
- Seems like it would be good to move the instructions about doing the login into the loop as well so that they only print if the user is being logged in.

WarpWing:
> This has some good elements and ideas in it. However, it doesn't seem to work correctly. A few thoughts here:
> 
>     * The `gh api user` command return 0 when logged in and non-zero when not logged in.  So I think the condition on that needs to not use `!`.  So the loop should be "while not logged in" or `while gh api user`.
> 
>     * Then in the loop do the `gh auth login` inside the loop.
> 
>     * Seems like it would be good to move the instructions about doing the login into the loop as well so that they only print if the user is being logged in.

Thanks for the feedback, I've implemented your suggestions! I moved the `gh auth login` login inside and make the logic revolve around `gh api user`.

PR #410, opened 2024-12-13
fix(dev): test rerun of fd2-up.linux.bash without double run

### LINKED ISSUE #405: Remove need to run fd2-up.bash twice on linux/wsl
Currently the `fd2-up.bash` script must be run twice the first time FarmData2 is started on linux or Windows (with WSL).   This is a bit of an inconvenience and different than the other platforms.  So it would be nice to remove this necessity.

The necessity to run `fd2-up.bash` twice is that the `fd2-up.linux.bash` sub-script uses the `newgrp` command to add a group.  This starts a new shell and thus terminates the script.  The `sg` command does not start a new shell and thus provides a way to modify the `fd2-up-linux.bash` script such that it will not have to be run twice.

Here is a snippet that is similar to what would be needed.  It adds two groups and leaves you in a `bash` shell where you user is in both groups:
```bash
#!/bin/bash

GROUP1=sggrp1
GROUP2=sggrp2

if [ -z "$IN_GROUP1" ]; then
  sudo groupadd "$GROUP1"
  sudo usermod -aG "$GROUP1" "$(whoami)"

  echo "First if"
  echo "My groups are...$(groups)"

  export IN_GROUP1=1
  exec sg "$GROUP1" "exec '$0' $(printf "'%s' " "$@")"

elif [ -z "$IN_GROUP2" ]; then

  sudo groupadd "$GROUP2"
  sudo usermod -aG "$GROUP2" "$(whoami)"

  echo "Second if"
  echo "My groups are...$(groups)"

  export IN_GROUP2=1
  exec sg "$GROUP2" "exec '$0' $(printf "'%s' " "$@")"

fi

echo "After if"
echo "My groups are...$(groups)"
newgrp "fd2grp"
```

WarpWing:
**Pull Request Description**

Closes #405 

Achieves this by adding a GROUPS_PROCESSED, added sg as opposed to newgrp moved the directory permission operations after the group setup, Added proper quoting for the script arguments in the sg command and created the .fd2/gids directory and saved GIDs before switching groups. This should allow for a single path. Would like thoughts (and a second test, this is hypothetical right now)

Main `sg` functionality looks like:

```bash
exec sg docker "sg fd2grp \"$0\" \"$@\""
```

**Licensing Certification**


WarpWing:
I would like to note that this is something I'd like to put for review but still working on (this is still untested) on setting up an environment for testing (I imagine guidelines would just be to setup a test env and if something goes wrong, inspect error logs for permission or group failures in assignment or permissions needed).  Would like some thoughts before proceeding.

braughtg:
Superseded by #410 which builds on the ideas here in a simplified way.

Issue #33, opened 2023-11-09
Improve efficiency of `lintstaged` that is run in git pre-commit hook


braughtg:
Currently if a `.vue` file (i.e. a component or entry point) changes all of its tests are also run.  If any of the tests for the component have also changed then they are run a second time.  So, if a `.vue` file is changed, we can skip running any of the changed tests because they will have already been run.  

A similar inefficiency occurs with changes to any of the library files (in `library` or in `lib.js` within the modules).  These inefficiencies should be addressed as well.

Issue #32, opened 2023-11-09
Have `pre-commit` git hook re-add staged files


braughtg:
When the `pre-commit` git hook fails and you go make a correction the changed files need to be staged again.  it is easy to forget to do that extra `git stage`.  The `pre-commit` hook should check if any of the staged files have been modified and then ask if they should be staged again.

PR #412, opened 2024-12-15
feat(dev): Added restage, linting rules and modified lintstaged for git pre-hook.

### LINKED ISSUE #19: Add linting rules to check test content
There are some additional linting checks that should be included:

- [ ] fix(dev): check no `.only` in test files
  - Linting during the `pre-commit` git hook should check any `.cy.js` files being committed to ensure that they do not contain a `.only` on a test.  The `pre-commit` hook should fail if any `cy.js` file contains a `.only`.
- [ ] fix(dev): check no `console.log` in test files
  - Linting during the `pre-commit` git hook should check any `.cy.js` files being committed to ensure that they do not contain `console.log` statements.  The `pre-commit` hook should fail if any `cy.js` file contains a `console.log` statement.

WarpWing:
**Pull Request Description**

Closes #19, #32, #33 

---

**Licensing Certification**

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -34,6 +34,47 @@ check_modified_staged_files() {
   return 1  # No modified staged files found
 }
 
+# Function to check Cypress test files for .only
+check_test_only() {
+  local files=$(git diff --cached --name-only --diff-filter=d | grep '\.cy\.js$' || true)
+  if [ -n "$files" ]; then
+    echo "Checking for .only in Cypress test files..."
+    for file in $files; do
+      if grep -n "\.only" "$file" > /dev/null; then
+        echo "Error: Found .only in $file"
+        echo "Please remove .only before committing."
+        return 1
+      fi
+    done
+  fi
+  return 0
+}
+
+# Function to check Cypress test files for console.log
+check_console_log() {
+  local files=$(git diff --cached --name-only --diff-filter=d | grep '\.cy\.js$' || true)
+  if [ -n "$files" ]; then
+    echo "Checking for console.log in Cypress test files..."
+    for file in $files; do
+      if grep -n "console\.log" "$file" > /dev/null; then
+        echo "Error: Found console.log in $file"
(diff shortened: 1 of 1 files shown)


braughtg:
Your approach to #32, by adding code to the pre-commit hook looks like it could work.  I think the thing to do would be to strip this PR down to dealing just with #32.  

Then if you would like to work on #19 and #33 open separate PRs for them.  But note that I don't think that the pre-commit script itself is the right place to solve #19 and #33.  I think the specific requirements of #19 and #33 are better addressed in other ways.  in particular: 
- For #19
  - The `eslint-plugin-cypress` module that is already in the project has the ability to check for `.only`. So if we can ensure that that rule is used on test files that should do the trick. 
  - Similarly `eslint` has a `no-console` rule that could be applied to test files.
- For #33 
  - The determination of which tests to run should probably be done in `.lintstagedrc.cjs`.

WarpWing:
> Your approach to #32, by adding code to the pre-commit hook looks like it could work. I think the thing to do would be to strip this PR down to dealing just with #32.
> 
> Then if you would like to work on #19 and #33 open separate PRs for them. But note that I don't think that the pre-commit script itself is the right place to solve #19 and #33. I think the specific requirements of #19 and #33 are better addressed in other ways. in particular:
> 
>     * For [Add linting rules to run during pre-commit git hook #19](https://github.com/FarmData2/FarmData2/issues/19)
>       
>       * The `eslint-plugin-cypress` module that is already in the project has the ability to check for `.only`. So if we can ensure that that rule is used on test files that should do the trick.
>       * Similarly `eslint` has a `no-console` rule that could be applied to test files.
> 
>     * For [Improve efficiency of `lintstaged` that is run in git pre-commit hook #33](https://github.com/FarmData2/FarmData2/issues/33)
>       
>       * The determination of which tests to run should probably be done in `.lintstagedrc.cjs`.

Alright, I'll close this branch and patch changes in a separate PR.

PR #421, opened 2025-02-13
feat(fd2): add disable support for MultiSelectorBase

### LINKED ISSUE #164: EquipmentSelector: Disallow duplicate equipment

RolandLocke:
**Pull Request Description**

### Changes:

- `MultiSelectorBase` converts the `options` prop into a `optionsObjects` array using the object format for the `options` prop in `SelectorBase`.
- `MultiSelectorBase` has a new prop: `allowDuplicateSelections`. This prop will allow duplicate options to be selected in the `MultiSelectorBase`'s `SelectorBase` components.  
   - A button has been added to `MultiSelectorBase` example page to demonstrate `allowDuplicateSelections`.
- When the `MultiSelectorBase` receives an `update:selected` event from one of its `SelectorBase` components:
  - If a new item has been selected, then the `MultiSelectorBase` will update its `optionsObjects` array by setting the `disabled` attribute of the object associated with the new selection to `true`.
- `MultiSelectorBase` passes an individualized `optionsObjects` list to each `selectorBase` component where the currently `selected` item is not disabled.
- Removed `keepDisabledSelected` property from `selectorBase` and all mentions from test files and example pages. This is no longer needed due to the changes above.

#### **Behavior Tests Added:**

1. **Check Appropriate Items Appear Disabled When Updating Selected**  
   - Ensures the correct items are disabled when the selection changes.

2. **Check Display Correct Disabled Status When New Item Selected**  
   - Verifies that selecting a new item updates the disabled statuses correctly.

3. **Check Display Correct Disabled Status When Deleted**  
   - Confirms that removing a selected item updates the disabled statuses.

4. **Check Appropriate Items Appear Disabled When Updating Options**  
   - Ensures that changing the available options updates the disabled statuses.

5. **`allowDuplicateSelections` Allows Users to Select Multiple Items**  
   - Tests that multiple selectors can choose the same item when duplicates are allowed.

6. **Turning `allowDuplicateSelections` Off Removes All Duplicates**  
   - Verifies that disabling duplicate selections removes existing duplicates.

#### **Events Tests Added:**

7. **Emits `"update:selected"` When Duplicate Items Are Selected**  
   - Checks that selecting duplicate items triggers an update event.

8. **Emits `"update:selected"` When `allowDuplicateSelections` Is Off and Duplicate Items Are Selected**  
   - Ensures that turning off duplicate selections emits an update event with the correct values.

Closes #164 

---

**Licensing Certification**


components/MultiSelectorBase/MultiSelectorBase.vue
```diff
@@ -205,6 +206,11 @@ export default {
         this.valid[0] = !this.required;
       }
     },
+    disableSelectedItems() {
```
RolandLocke: @braughtg I'm encountering an error here, but I'm not sure why. We need this function to somehow iterate over all the items in this list both when selected is updated and when the component is created.

braughtg: This looks like it is disabling all of the options?  Should it only be disabling the ones that are currently selected?

components/MultiSelectorBase/MultiSelectorBase.vue
braughtg, line 1: Overall I wasn't sure I followed all of the logic behind the changes in this file.  Maybe drop by early this week and we can walk through it?

RolandLocke: Yes, We can walk through it. I didn't see this earlier.

RolandLocke: @braughtg Okay, I've made the changes we've discussed: 

- [ ] `AllowDuplicates `should not be included in a ternary statement
- [ ] if `AllowDuplicates `is changed from true to false, update the selected items
- [ ] Move `processedOptions `into a "normally" called method

Let me know what you think before I move forwards.

I'm also seeing that the first test in the `MultiSelectorBase.behavior` is still failing (a number of them are). Maybe there was something else we changed in the `selectorBase` component that fixed this issue? I'm sure we've dealt with this before.

RolandLocke:
@braughtg I updated all the tests where we saw the following behavior:

Items initially passed to `Selected` did not populate the component properly. Items populated the dropdowns in the example page as expected, the tests would not pass. Instead, initial selected items in all tests are now passed via wrapper (prop update). 

Lastly, I removed the test `Option selection disables elements in selectorBases` that I had previously created to test new features. This was a test that I created.

RolandLocke:
> Items initially passed to `Selected` did not populate the component properly. Items populated the dropdowns in the example page as expected, the tests would not pass. Instead, initial selected items in all tests are now passed via wrapper (prop update).

Again, this test behavior is weird, and you should check this out before we move forwards.

braughtg:
@RolandLocke I think I have a handle on this but there have been some rippling changes.  I think I'll have it all patched tomorrow and will push some changes to this PR then.

.fd2/gids/docker.gid
braughtg, line 1: Remove this file from the PR.  It was added to your repo when we tried the codespaces dev environment.  It will be added to the `.gitignore` when that stuff is merged.

.fd2/gids/fd2grp.gid
braughtg, line 1: Remove this file from the PR.  It was added to your repo when we tried the codespaces dev environment.  It will be added to the `.gitignore` when that stuff is merged.

braughtg (changes requested):
See my more extensive comment below (https://github.com/FarmData2/FarmData2/pull/421#issuecomment-2697978733)

braughtg:
One thing that is a little strange is that all of the tests pass on `development` which has the changes to the `SelectorBase` already merged. But in this PR branch things that use `SelectorBase` begin to have failing tests (e.g. `CropSelector`, `TraySizeSelector`. I'm not sure what happened but that may be something worth diagnosing.  Note: There are other tests failing due to changes in `MultiSelectorBase` as well, but I think we should delay looking at those until we resolve the ones related to `SelectorBase`.

It looks as if this has to do with the line `immediate: true` that was added to the `watch` on `options`.  I think that the issue is that by adding `immediate` here the `checkSelectedOption` method is invoked immediately.  That method will clear the selection if the `selectedOption` does not appear in the `optionsObjects`.  But some components that use the `SelectorBase` require an API call to fetch the options.  So what happens is that the `selectedOption` is cleared before the `optionsList` gets set.

I think one possible solution is to adjust `SelectorBase` so that:
- `checkSelectedOption` only does its thing if the list of `optionsObjects` is not empty (or is only called when `optionsObjects` is not empty - consider what is more readable/maintainable).
- `checkSelectedOption` is invoked any time:
  -  `optionsOpbjects` changes.
  - `selected` changes.

Currently, there seems to be some overlap in what `checkSelectedOption` is doing and what the `watch` on `selected` is doing.  I wonder if we can factor that out to help simplify the logic?

.fd2/gids/docker.gid
RolandLocke: I didn't catch these, I thought I had ran the code spaces environment on a different branch...

braughtg: We did use another branch for that test. But it created these files that were not committed to that branch. So when you switched back to your branch they looked like changes to git and it added them when you did an `add .`.  No worries.  Thanks for pulling them out.

RolandLocke:
This hasn't been flipped to review as I have found a couple of failing e2e tests. Specifically direct_Seeding.submitReset tests are failing for a number of components.

Issue #420, opened 2025-02-05
Refactor `soil_disturbance` entrypoint to use  the `ActivePlantAssetPicklist` Component


braughtg:
The `ActivePlantAssetPicklist` component will be created by #419.  Thus, #419 must be completed before this issue.

When #419 is complete the  `soil_disturbance` entrypoint should be refactored to remove the `PicklistBase` and use an `ActivePlantAssetPicklist` component instead.  

- All of the code related to fetching and formatting the list of active plant assets should be removed from the `soil_disturbance` entrypoint and be replaced by the use of the `ActivePlantAssetPicklist`.
- All of the test code for tests that duplicate component tests on the `ActivePlantAssetPicklist` component should be removed from the `soil_disturbance` entrypoint's tests.
  - e.g. checking that the correct list of plant assets is loaded for a location. However there should still be tests for things like changing the location changes the list, and that the correct props are passed to the component.

PR #422, opened 2025-02-18
feat(comp): Adds `ActivePlantAssetPicklist` Component for Selecting Active Plantings by Location

### LINKED ISSUE #419: Factor `ActivePlantAssetPicklist` Component out of `SoilDisturbance` entrypoint
The `SoilDisturbance` entrypoint contains a `PicklistBase` that is used to display all of the active plantings in the selected location (circled below):

Image

The functionality of this `PicklistBase` should be factored out into an `ActivePlantAssetPicklist` which:
- accepts a location name  (e.g. ALF) as a prop.
- automatically populates with the active plant assets that exist in that location.
- provides appropriate _pass through_ props to `PicklistBase`
- emits the same events as the `PicklistBase`

In addition:
- component tests should be created for the `ActivePlantAssetPicklist` component.
- an example page should be created for the `ActivePlantAssetPicklist` component.

Shahir-47:
**Pull Request Description**

This PR adds the `ActivePlantAssetPicklist` component, which allows users to view and select active plantings in a given location. The `ActivePlantAssetPicklist` component:  

- Accepts a location name as a prop (e.g., `ALF`).  
- Displays all active plantings in the selected location.  
- Allows users to pick crops from the available plantings.  
- Passes relevant props to `PicklistBase`.  

Additionally:  
- Added tests for `ActivePlantAssetPicklist`.  
- Created an example page for `ActivePlantAssetPicklist`.  

Closes #419 

---

**Licensing Certification**


Shahir-47:
@braughtg This PR is ready for review.

components/PicklistBase/PicklistBase.vue
```diff
@@ -645,13 +645,6 @@ export default {
 
         // Rebuild pickedRows from picked Map
         this.pickedRows = new Array(this.sortedRows.length).fill(0);
-        for (let i = 0; i < this.sortedRows.length; i++) {
```
Shahir-47: I removed this because I don't want the checkbox selections to be preserved when the `rows` prop changes. Previously, it kept the selected checkboxes, but now, when the parent component (like `ActivePlantAssetPicklist`) updates the `rows` prop, the checkbox selections should no longer be retained.

braughtg: What was the motivation for this change?

components/ActivePlantAssetPicklist/ActivePlantAssetPicklist.vue
```diff
@@ -0,0 +1,228 @@
...
+  methods: {
+    handleUpdatePicked(event) {
+      /**
+       * Emitted when the picked crops have changed.
+       *
+       * @event update:picked
+       * @property {Map<number, Object>} event - A Map where the keys are the indices of the picked rows in the `rows` prop of picklistBase, and the values are objects representing the picked rows and their data.
+       *
+       */
+      console.log(`called picked`);
```
braughtg: Be sure to get rid of any of these added for debugging.

This one also prevents the docs for the `update:picked` event from being generated because it comes between the comment and the `emit`.

components/PicklistBase/PicklistBase.vue
Shahir-47: When switching locations in the `ActivePlantAssetPicklist` component, I noticed an issue with checkbox selection. If the first row's checkbox is selected before switching locations, it remains checked after the update. The removed code caused this by carrying over the previous selections when the `row` prop updated due to the location change.

To prevent unintended selections, I suggest resetting the selected checkboxes when switching locations. This issue also occurs in other components/pages using `PicklistBase`, such as `TransplantPicklist` and the `SoilDisturbance` page.

modules/farm_fd2_examples/src/entrypoints/picklist_base/App.vue
braughtg, line 1: I noticed that the `update:picked` event payload is processed by this example page, so it appears differently than it does in your example for the `ActivePlantAssetPicklist`.  It would make sense to modify this example so that it displays the `update:picked` event payload in its raw form as you do in the `ActivePlantAssetPicklist` example.

components/PicklistBase/PicklistBase.vue
braughtg: I agree, resetting the selected rows in an `ActivePlantAssetPicklist` makes good sense when the location changes. But I don't think the selections should be reset any time `rows` changes (e.g. if the page using the component were to add or remove a row.). In that case, the selection can remain.

So, I think we should revert the changes to `PicklistBase` and put the behavior that resets the selections when the location changes into `ActivePlantAssetPicklist` instead of in `PicklistBase`.

You'll want to have two different tests. One to ensure that the selections stay when `rows` changes (this test should be added to the `PicklistBase` tests - because it must not be there now or it would have failed). The other to ensure that the selections are reset when the location changes (which will need to be tested in `ActivePlantAssetPicklist`.

Shahir-47: I originally implemented this assuming that sorting the rows would update the `rows` prop, so I may have intended to preserve the selection when `rows` changed. However, sorting does not affect the `rows` prop because it is handled internally using the local `sortedRows` array, which does not modify the `rows` prop at all.

The only time the `rows` prop watcher is triggered is when a new row is added or an existing row is deleted. The code inside the watcher ensures that the new data is displayed in the user's selected sorting order.  

For example, if I choose location `ALF` and set the `PicklistBase` table to sort ascendingly by bed names, then switch to location `CHUAU`, the watcher ensures that the new data is also displayed in ascending order by bed names, just as it did for `ALF`.

modules/farm_fd2_examples/src/entrypoints/active_plant_asset_picklist/active_plant_asset_picklist.exists.e2e.cy.js
```diff
@@ -0,0 +1,10 @@
+describe('Check that the active_plant_asset_picklist entry point in farm_fd2_examples exists.', () => {
+  it('Check that the page loaded.', () => {
+    // Login if running in live farmOS.
+    cy.login('admin', 'admin');
+    // Go to the main page.
+    cy.visit('fd2_examples/comment_box/');
```
braughtg: Adjust to check for this example page.

components/ActivePlantAssetPicklist/ActivePlantAssetPicklist.vue
```diff
@@ -0,0 +1,228 @@
...
+            this.picklistLabels = {
+              crop: 'Crop',
+              bed: 'Bed',
+              timestamp: 'Planted Date',
+            };
+          }
+        } catch (error) {
+          console.error('Error fetching plant assets:', error);
+          this.affectedPlants = [];
+          this.$emit('error', 'Unable to fetch plant assets.');
```
braughtg: The `error` event should have a doc string with a description.

braughtg (changes requested):
Overall, looking good.  I left a few comments to review and discuss.  I did not fully review the tests because I want to wait until we think the implementation is final because changes will likely ripple through the tests.

Be sure to also update the `component_examples/App.vue` to include a link to the new example.

Shahir-47:
@braughtg, I wasn't able to test if `PicklistBase` checkboxes remain picked if new rows are added to the `rows` prop due to the Cypress limitation I mentioned earlier. So I have left that for now and also removed a test from `ActivePlantAssetPicklist` that checked if picked selections were reset if the location changes. However, for similar reasons to the `PicklistBase` test, I believe this only works because Cypress resets the selections whenever props are changed, regardless of the component's reset logic.

So, the tests are passing for that reason and not necessarily due to the component's intended reset behavior. 

Apart from these two issues, I have made the changes you requested, and this PR is ready to be reviewed.

Shahir-47:
@braughtg,  

Looking at [Issue #420](https://github.com/FarmData2/FarmData2/issues/420), I’m considering adding:  
1. A **computed property** `plantsAtLocation`  
2. A **new method** to calculate the area based on picked beds (similar to how it’s done in Soil Disturbance).  

### **Why This is Needed**  
Without these additions in `ActivePlantAssetPicklist`, the Soil Disturbance page will need to make an **extra call** to `farmosUtil.getPlantAssets` to determine if plants exist at a location. This results in two separate calls:  
- **One inside `ActivePlantAssetPicklist`** to populate the table.  
- **One in the Soil Disturbance page** to check for plants at the location.  

This second call is necessary because `LocationSelector` relies on `plantsAtLocation` to determine whether to:  
```vue
v-bind:allowBedSelection="!plantsAtLocation"
```
Similarly, the **termination div** visibility depends on it:  
```vue
<div
  id="termination-event-group"
  data-cy="termination-event-group"
  class="d-flex flex-column align-items-center"
  v-if="plantsAtLocation"
>
```
Additionally, the Soil Disturbance page **needs** to call `farmosUtil.getPlantAssets` to calculate the area field in the Soil Disturbance page.  

### **Proposed Improvement**  
If `ActivePlantAssetPicklist` handles both the area calculation and the `plantsAtLocation` check, it consolidates everything into **one API call** instead of two. This would improve efficiency and simplify logic.

Shahir-47:
@braughtg I made some updates to `ActivePlantAssetPicklist` by adding two new emits:

1. **`update:area`** - This calculates and emits the area percentage based on fully selected beds.
2. **`hasPlants`** - This emits `true` if the location has active plant assets and `false` otherwise.

I also wrote corresponding tests in [`ActivePlantAssetPicklist.events.comp.cy.js`](https://github.com/FarmData2/FarmData2/blob/cc0564c944fdceaf94066182ddb370263acad01b/components/ActivePlantAssetPicklist/ActivePlantAssetPicklist.events.comp.cy.js) to verify these emits and updated the example page accordingly.

These changes were made **alongside** edits to the Soil Disturbance page, but I **did not push** the Soil Disturbance changes to keep this PR clean. If you'd like to see how this impacts the Soil Disturbance page, please refer to the PR #425. However, that PR does **not** include `ActivePlantAssetPicklist`.

If you'd like to see both changes together, you can clone my branch [`refactor_soil_dist`](https://github.com/Shahir-47/FarmData2/tree/refactor_soil_dist) from my fork, where both the Soil Disturbance and Active Plant Asset changes are implemented together.

Let me know if you have any questions!

braughtg:
> 1. **`update:area`** - This calculates and emits the area percentage based on fully selected beds.

I'm not quite understanding what the area calculation is computing.  For example, when I pick "ALF":
- Before I have selected a location, the `update:area` event has emitted 100.
- When I have picked a location, but have no plants checked in the list it is emits 100.
- If I check one of the plants in ALF-1 it emits 100.
- If I check both of the plants in ALF-1 it emits 50.
- If I check all three plants in ALF-1 and ALF-2 it emits 100.

~~I am wondering if the area computation needs to be done for each of the picked plants.  For example, if there are 2 plants in "ALF-1" and I pick one of them, the structure emitted by `update:picked` would include `area: 50`.  If I then pick the second plant it too would be emitted with `area: 50`.  If I pick the one plant in ALF-2 then it should be emitted with `area: 100`.~~

~~This will also mean that the `SoilDisturbance` page needs an update... For example if 50% of ALF-1 is affected and 100% of ALF-2 is affected that would need to be reflected in the logs, which would require creating 2 different logs.~~

That wasn't right... we need the percentage of the *field* that was affected.  So the computations should be:
- If one of the plants in ALF-1 is picked it should emit 25% (1/2 of 1/2 of ALF).
- If both plants in ALF-1 are picked it should emit 50%.
- If only the plant in ALF-2 is picked it should emit 50%.
- If all three plants are picked it should emit 100%.

Shahir-47:
@braughtg All tests for ActivePlantAssetPicklist are passing.

Since there were so many tests for `update:area`, I wrote them in a separate test file to keep things organized. The area calculation needed a lot of tests because there are three different types of locations, and each required different logic to determine the area:

- **Locations with beds:**  
  The area is calculated using the summation formula we discussed. For each bed, we take the number of picked crops, divide it by the total crops in that bed, sum up these fractions for all beds, and then divide by the total number of unique beds to get a percentage.

- **Locations with active plant assets but no beds:**  
  Since there are no beds to divide by, we split the area evenly across all plant assets. If half the plants are picked, the area is 50%. If all are picked, the area is 100%.

- **Locations with no active plant assets:**  
  The area is set to 100% since there are no crops to pick.

### About the Tests:
- The tests cover all possible scenarios for `update:area`:
  - **Initial state:** Checking that `update:area` emits `0` when the component is created.
  - **Selecting and deselecting crops:** Making sure the calculation updates correctly for different types of locations.
  - **Switching between locations:** Ensuring the correct reset behavior occurs when changing locations.
  - **Ensuring `update:area` does not emit unnecessarily:** If no crops are picked and the location changes, `update:area` should not emit.


Let me know if you have any feedback or if anything should be changed!

Shahir-47:
@braughtg This PR is ready for review!

PR #425, opened 2025-03-04
refactor(fd2): soil dist page for ActivePlantAsset component

### LINKED ISSUE #420: Refactor `soil_disturbance` entrypoint to use  the `ActivePlantAssetPicklist` Component

Shahir-47:
**Pull Request Description**

This PR refactors the `soil_disturbance` entrypoint to use the `ActivePlantAssetPicklist` component instead of `PicklistBase`. 

### **Changes Made:**
- **Replaced `PicklistBase` with `ActivePlantAssetPicklist`** in the `soil_disturbance` entrypoint.
- **Removed redundant code** from `soil_disturbance` that fetched and formatted plant assets, as this logic is now handled by `ActivePlantAssetPicklist`.
- Area calculation is now handled within `ActivePlantAssetPicklist`.
- The LocationSelector's `BedPicker` and `ActivePlantAssetPicklist` visibility are now controlled via events emitted from  `ActivePlantAssetPicklist`

This refactor aligns with issue **[#420](https://github.com/FarmData2/FarmData2/issues/420)** and **depends on the completion of [#419](https://github.com/FarmData2/FarmData2/issues/419)**.

**Closes #420**

---

**Licensing Certification**


Shahir-47:
@braughtg This PR contains edits to the `soil_disturbance` page based on the implementation of `ActivePlantAssetPicklist` in **PR #422**.

The refactor ensures that:
- `PicklistBase` is replaced with `ActivePlantAssetPicklist`.
- Redundant logic for fetching and formatting plant assets is removed, as `ActivePlantAssetPicklist` now handles this.
- The **area calculation** and the visibility of the **Bed Picker (`LocationSelector`) and `ActivePlantAssetPicklistBase`** are now managed via emitted events from `ActivePlantAssetPicklist`, simplifying the code.

For a complete understanding of these changes, **refer to PR #422**, which introduces the `ActivePlantAssetPicklist` component.

Let me know if you have any questions!

Shahir-47:
@braughtg,

The `soil_disturbance` page and its tests have been updated and are ready for review.

- `PicklistBase` has been replaced with `ActivePlantAssetPicklist`.
- Redundant logic for fetching/formatting plant assets has been removed, as this is now handled internally by the new component.
- Tests were refactored to remove those that duplicated component behavior. I mainly updated the area computation test and removed checks on the `affectedPlants` array, since that's now encapsulated within the component.
- Updated visibility checks in the tests from `.should('not.exist')` to `.should('not.be.visible')` to reflect the switch from `v-if` to `v-show`.

modules/farm_fd2/src/entrypoints/soil_disturbance/soil_disturbance.equipment.e2e.cy.js
```diff
@@ -103,78 +104,108 @@ describe('Soil Disturbance: Equipment Component', () => {
       .find('input')
       .eq(0)
       .should('be.checked')
+      .wait(1000)
```
Shahir-47, line 111: Had to add a short wait here due to timing issues with the checkbox not being reliably clickable right away. Tried other Cypress strategies (like retrying clicks, assertions, and `.should('exist')`) but none worked consistently. Suppressing the lint warning with a comment above.

modules/farm_fd2/src/entrypoints/soil_disturbance/soil_disturbance.equipment.e2e.cy.js
```diff
@@ -103,78 +104,108 @@ describe('Soil Disturbance: Equipment Component', () => {
...
       .select('G');
     cy.get('[data-cy="soil-disturbance-equipment-form"]')
       .find('[data-cy="soil-disturbance-area"]')
       .find('[data-cy="numeric-input"]')
-      .should('have.value', 100);
+      .should('have.value', 1);
 
-    cy.get('[data-cy="picklist-checkbox-0"]').check();
+    // Does not check the checkbox without the wait method
+    cy.get('[data-cy="picklist-checkbox-0"]').wait(1000).check();
```
Shahir-47, line 192: Same reason as above — the checkbox doesn’t reliably become interactive without a short wait. Other Cypress retry strategies didn’t resolve it consistently.

modules/farm_fd2/src/entrypoints/soil_disturbance/soil_disturbance.equipment.e2e.cy.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable cypress/no-unnecessary-waiting */
```
Shahir-47: Used this to suppress the linting issue related to `wait()` usage, since it was necessary for consistent checkbox interaction during tests.

braughtg: Use `/* eslint-disable-next-line ... */` and place it just before each line with a wait.  That way if any other waits are added to the file they will still be flagged.  This ensures that the use of wait is a conscious decision each time it is used.

braughtg:
Here are a few other things that we will need to consider with this form:
- A termination event should probably require that at least one plant asset is selected. I think making the `ActivePlantAssetPickList` a required element if termination event is checked should do this.
- Currently it is possible to create a soil disturbance in an location but not select an active plant (or a bed).  What does that mean? Should this not be allowed if the location has plant assets or beds?  If the location does not have any plant assets or beds then it seems to make sense that the disturbance would apply to the whole (100%) field.
- When a location has both beds and active plant assets, if there is a bed without active plant assets, we don't currently have a way to do a soil disturbance on that bed.  Maybe we can add a row to the table for empty beds (w/ Crop and Planted Date) left blank? Or maybe there is another solution?

braughtg:
This looks good overall.  When I `runAllTests.bash` I am getting a failure:

```
Running:  soil_disturbance/soil_disturbance.equipment.e2e.cy.js                         (50 of 55)


  Soil Disturbance: Equipment Component
    ✓ Soil disturbance equipment header exists, is visible (1107ms)
    ✓ Soil disturbance equipment body exists, is visible (396ms)
    ✓ Validity styling shown in soil disturbance component (534ms)
    ✓ Proper numeric fields appear when equipment is selected (597ms)
    ✓ Equipment and numeric fields are required (597ms)
    1) Updating beds updates area


  5 passing (16s)
  1 failing

  1) Soil Disturbance: Equipment Component
       Updating beds updates area:

      Timed out retrying after 10000ms
      + expected - actual

      -'7'
      +'25'
      
      at Context.eval (http://farmos/__cypress/tests?p=src/entrypoints/soil_disturbance/soil_disturbance.equipment.e2e.cy.js:56:137)
```

Note: This also happens when I just run the e2e tests with `test.bash --e2e --live --fd2`

Shahir-47:
@braughtg I believe one of the earlier tests is modifying the database, which causes this test to fail. After adding `cy.task('initDB')` and running `runAllTests.bash`, all tests now pass. This PR is ready for review.

PR #430, opened 2025-04-02
fix(fd2): SoilDisturbance input form improvements

### LINKED ISSUE #429: feat(fd2): SoilDisturbance Updates
There are a few updates that need to be made to the `SoilDisturbance` form:

- [ ] - When the "Termination event"check box is checked, we should require that at least one row in the `ActivePlantAssetPicklist` to be selected. I think that making the `ActivePlantAssetPicklist` a required element when "Termination event" is checked should do this. When `ActivePlantPicklist` is required its status will be invalid unless at least one row is selected.

- [ ] - It is possible to create a soil disturbance in an location with neither a row in the `ActivePlantAssetPicklist` selected nor a bed selected (if there are no active plant assets in the location).  
  - I think this actually makes sense, because we want to allow for a situation where no bed has been defined for part of a location.  In that situation, a soil disturbance will affect no active plants, no beds, but some % of the location.  In this situation, it is the user must set the area percent appropriately.
  - Note: If the selected location has neither active plant assets nor beds then any soil disturbance should apply to the whole (100%) location. In this case the area should default to 100%.

- [ ] - If a location has both beds and active plant assets, and there is an empty bed that does not currently contain any active plant assets, we need a way to record a soil disturbance on the empty bed. Currently because there are active plant assets in the location the `ActivePlantAssetPicklist` is displayed and there is no way to choose an empty bed for the soil disturbance.
  - One possibility is to show both the `ActivePlantAssetPicklist` and the `BedPicker` within the `LocationSelector` when a location has both active plants and beds. With this solution, it will be necessary to synchronize the `ActivePlantAsssetPicklist` with the `BedPicker`.  That is, if a bed is checked, then all rows of the `ActivePlantAssetPicklist` for that bed are also checked.  Similarly, if all rows in the `ActivePlantAssetPicklist` for a bed are checked, then the bed should also be checked.

Shahir-47:
**Pull Request Description**

This pull request addresses improvements to the `SoilDisturbance` form as outlined in the corresponding issue:

- When the "Termination event" checkbox is checked, the `ActivePlantAssetPicklist` component now becomes required. This ensures users select at least one active plant asset for termination events, preventing accidental submissions without specifying affected plants.

- The form now properly allows creating a soil disturbance event in a location that has neither active plant assets nor beds. In such cases, the disturbance applies to the entire location (defaulting to 100% area), requiring the user to explicitly set the appropriate percentage if different from 100%.

The synchronization of bed selection with the ActivePlantAssetPicklist for locations containing both beds and active plant assets is not covered by this PR yet.

Closes #429 

---

**Licensing Certification**

FarmData2 is a [[Free Cultural Work](https://freedomdefined.org/Definition)](https://freedomdefined.org/Definition) and all accepted contributions are licensed as described in the LICENSE.md file. This requires that the contributor holds the rights to do so. By submitting this pull request **I certify that I satisfy the terms of the [[Developer Certificate of Origin](https://developercertificate.org/)](https://developercertificate.org/)** for its contents.


Shahir-47:
@braughtg This PR addresses the first two points from #429 and is ready for your review.

Regarding the third point, as mentioned [here](https://github.com/FarmData2/FarmData2/issues/429#issuecomment-2771866953), we don't have direct access to the full row data of the `ActivePlantAssetPicklist` in the `soil_disturbance` page. All row data is handled internally within the `ActivePlantAssetPicklist` component, which makes synchronizing bed selections with picklist rows non-trivial.

I'm open to discussing a possible solution for this point further.

braughtg:
A few comments:
- Looks like the `BedSelector` is coming up as expected.  However, it is appearing with all beds selected.  I think it should probably have no beds selected by default in this case.

- I can think of a few possible solutions to the issue you mention about the third point.
  - We could add a `beds` prop to the `ActivePlantAssetPicklist` that would indicate which beds are selected in the `LocationSelector`.  The `ActivePlantAssetPicklist` could then also emit a `beds` event that would indicate which if any beds are fully selected within the picklist.
  - We could add a `BedPicker` component to the `ActivePlantAssetPicklist` component and hide the one associated with the `LocationSelector`.  This way, the `ActivePlantAssetPicklist` could do the synchronization internally.

Of these two options for the third point, I'm leaning toward the 2nd option, but would like to discuss it to see if you foresee any issues.

Shahir-47:
@braughtg

I was able to synchronize the ActivePlantAssetPicklist with the BedPicker. There are two problems though:

1. When I uncheck a bed from the BedPicker I cannot uncheck all the rows from the PicklistBase. I could add logic to do that but it creates another problem. For example, if I uncheck one row for bed "alf-1" (assuming all rows with the bed field "alf-1" were checked), the watcher immediately deselects the "alf-1" checkbox in the BedPicker. That then causes it to remove all the rows for that bed in the PicklistBase even though the intent was to uncheck only that single row.

2. When I press the All button in the `bedPicker` all beds are initially selected. Then the watcher for pickedRows is triggered and it modifies the BedPicker to only select beds that the PicklistBase has. Note that if you press All in the `bedPicker` twice it works as expected; the first press checks all beds present in the PicklistBase and the second press adds the remaining beds that are not in the PicklistBase.

Below are two flowcharts illustrating these scenarios.

### Scenario 1: Manual Row Uncheck in PicklistBase

```mermaid
flowchart TB
    A["Assume all rows for bed ALF-1 are initially checked (checkedBeds = [ALF-1]). Then the user manually unchecks one row for bed ALF-1 in PicklistBase."]
    A -- "Step 1: Triggers the PickedRow watcher" --> PR["PickedRow Watcher"]
    PR -- "Step 2: Detects a partial selection for bed ALF-1 and removes ALF-1 from checkedBeds (now checkedBeds = []), triggering the CheckedBeds watcher" --> CB["CheckedBeds Watcher"]
    CB -- "Step 3: Observes that checkedBeds is empty and unchecks all rows for bed ALF-1 in PicklistBase, which re-triggers the PickedRow watcher" --> PR
    PR -- "Step 4: Finds no new rows are selected; checkedBeds remains empty, so no further updates occur" --> End["End"]
```

### Scenario 2: Pressing the All Button in BedPicker

```mermaid
flowchart LR
    A[User presses All button in BedPicker]
    CB[CheckedBeds Watcher]
    PR[PickedRow Watcher]
    END[End]

    A --> CB
    CB -- "Step 1: Selects all rows from PicklistBase and triggers the PickedRow watcher" --> PR
    PR -- "Step 2: Recognizes that all rows are selected but filters checkedBeds to include only beds present in PicklistBase, triggering the CheckedBeds watcher" --> CB
    CB -- "Step 3: Updates the selection using only the beds from PicklistBase and triggers the PickedRow watcher again" --> PR
    PR -- "Step 4: Verifies that all rows are selected; since checkedBeds is already accurate, no further updates occur" --> END
```

I appreciate any input you have on these issues and welcome further suggestions to improve the synchronization.

braughtg:
> I appreciate any input you have on these issues and welcome further suggestions to improve the synchronization.

My first thought here is that much if this is due to using `v-model` on `picked` for the `BedPicker`.  That made good sense when the `BedPicker` was being used in the `LocationSelector`.  However, here I think you'll need to use a `v-on:update:picked` handler.  That will give you more control over what is updated and how than the `v-model` does.

Note that this is how you are handling the changes to `picked` in `PicklistBase` and it seems like all of the changes that happen in that direction are working well.

braughtg:
@Shahir-47 Can you give me a quick update on this, is it ready for review again after switching to a `v-bind` and `v-on`?

Shahir-47:
@braughtg I’ve pushed the latest version of `ActivePlantAssetPicklist.vue`. The switch to `v-bind` and `v-on` is working well so far.

I haven’t pushed the tests yet since I’m still working through them and brainstorming a few edge cases. Just wanted to confirm if the current behavior makes sense to you before I finalize everything:

- When beds are selected, the picklist updates accordingly.
- When an item is selected from the picklist, the bed picker also updates to reflect that.
- The "All" button is now working as expected in both the bed picker and the picklist.
- If a picklist row is deselected, that bed will be deselected in the bed picker.
- Note that deselecting a checkbox in the bed picker will not automatically deselect all the rows corresponding to that bed.

Also, for manual testing, I’ve been using the example page instead of the Soil Disturbance page. You can check it there to see the updated behavior in action.

Does this behavior look good to you? Or should I add any new props or make something required to improve clarity or consistency?

Let me know your thoughts. I’ll keep working on the tests in the meantime.

braughtg:
> The switch to `v-bind` and `v-on` is working well so far.

Good glad to hear it.  

I haven't yet had the time to pull and run the code, but I think the behavior should reflect that "a bed appears checked when the operation will affect that bed." 

This would imply a few changes to how you described the behavior above:

> * If a picklist row is deselected, that bed will be deselected in the bed picker.

The bed should only be automatically deselected if the last row for that bed is manually deselected.

> * Note that deselecting a checkbox in the bed picker will not automatically deselect all the rows corresponding to that bed.

If a bed is manually deselected, then all rows for that bed should be automatically deselected.  Conversely, if a bed that was not selected is manually selected, then all rows for that bed should be automatically selected.

Shahir-47:
@braughtg

Thanks for the clarification on the expected behavior.

Based on this, I assume the line:

> “The bed should only be automatically deselected if the last row for that bed is manually deselected”

means that as long as I have at least one row selected for a given bed, the corresponding bed checkbox should be checked.

And based on this, I assume:

> “If a bed that was not selected is manually selected, then all rows for that bed should be automatically selected”

means that when a bed is checked, it should select all rows associated with it.

The issue I’m running into is that when I select a single row, the bed checkbox becomes checked to reflect that. But then, because that checkbox is now marked as checked, it triggers the logic that says "all rows for this bed should be selected," which ends up selecting everything even though I only selected one row to begin with. This creates a loop where one row selection leads to all rows being selected.

Let me know what you think would be the best way to handle this.

braughtg:
> > “The bed should only be automatically deselected if the last row for that bed is manually deselected”
> 
> means that as long as I have at least one row selected for a given bed, the corresponding bed checkbox should be checked.

Correct.

> And based on this, I assume:
> 
> > “If a bed that was not selected is manually selected, then all rows for that bed should be automatically selected”
> 
> means that when a bed is checked, it should select all rows associated with it.

Correct.

> 
> The issue I’m running into is that when I select a single row, the bed checkbox becomes checked to reflect that. But then, because that checkbox is now marked as checked, it triggers the logic that says "all rows for this bed should be selected," which ends up selecting everything even though I only selected one row to begin with. This creates a loop where one row selection leads to all rows being selected.
> 
> Let me know what you think would be the best way to handle this.

You'll need to build a way to distinguish between when a bed is checked manually and when it is checked because it wasn't previously and a row has been manually checked.  The way to do this will be to incorporate some state and logic into the component that allows you to distinguish these different scenarios and handle them differently.  You can probably reason this out.  But another approach would be to design a state machine that describes the desired behavior and then implement it in your component.

Shahir-47:
@braughtg thanks for your guidance! I implemented a solution using an `updateInProgress` flag that prevents the issue I described above.

**The problem scenario was:**
1. User selects a single row
2. This causes the bed checkbox to become checked (correct behavior)
3. The checked bed would trigger logic to select all rows for that bed
4. Result: All rows get selected when user only wanted one row

**My solution:**
When a user selects a row:
1. Set `updateInProgress = true` at start of handler
2. Update the row selection
3. Update the bed checkbox based on row selection
4. Set `updateInProgress = false` at end of handler

The key is that when the bed becomes checked in step 3, it emits an update event that would normally trigger `handleBedPickerUpdate`. But when that handler runs, it immediately encounters:
```js
if (this.updateInProgress) return;
```
This guard prevents the bed selection from triggering row selection logic during the same update cycle, preserving the user's original intent to select just one row.

Similarly, when a user selects a bed:
1. Set `updateInProgress = true` at start of handler
2. Update the bed selection
3. Update all corresponding rows based on the bed selection
4. Set `updateInProgress = false` at end of handler

During step 3, each row selection would normally trigger `handleUpdatePicked`, but the same guard prevents these events from cascading back to affect beds.

This approach creates a clear boundary around each user interaction, allowing exactly one synchronization pass per click without needing complex state tracking.

braughtg:
> thanks for your guidance! I implemented a solution using an `updateInProgress` flag that prevents the issue I described above.

I just ran the `ActivePlantAssetPicklist` example and it seems to be working correctly.  Nice work!

Now just need to update the tests for that component, and update the `SoilDisturbance` page and its tests.  There is one case that I think will be important to check that we get correct.  That is where there is mixture of beds with plant assets and without plant assets.  For example, if `ALF` is picked as the location and `ALF-1` and `ALF-3` are selected as beds.  In that case, there are plant assets in `ALF-1` but not in `ALF-3` so we'll need to be sure the correct soil disturbance logs get created.

Shahir-47:
@braughtg,

I refactored the `ActivePlantAssetPicklist` component to add full syncing behavior between the `BedPicker` and `PicklistBase`, and I wrote exhaustive tests covering every possible selection scenario, including edge cases like beds without rows and rows without beds. This PR is ready for review. Once you approve it, I can start integrating these changes into the Soil Disturbance page.

braughtg (changes requested):
This looks really good.  Great progress.  Though, I do see a few issues:

1. There is an case when the component is required and there is a bed with no active plant assets. To see it in the example page:
   1. Make the component required.
   2. Enable showValidityStyling
   3. Select ALF
   4. Choose ALF-3

   In this scenario, the `Picklist` is still indicating that a row must be selected. However there are no rows to be selected, because ALF-3 does not have any active plantings. I think we can simplify this to making the component valid anytime any row or any bed is picked.  The "any row" part of that accounts for a location with plant assets but no beds (e.g. G).

2. A related issue comes up when selecting a location that has beds but no active plantings.  To see this:
   1. Make the component required.
   2. Enable showValidityStyling
   3. Choose Location H

   In this scenario, there are no rows to select.  So I think we should hide the pick list here and the component should be valid if any bed is checked.  This really reduces to the same validity condition as in 1 above.

3. I'm getting some E2E test failures in `DirectSeeding` because some components are not visible.

4. There was an E2E failures in `SoilDisturance: Equipment Component`.  This is a known flaky test.  It would be great to patch it when you make the updates to `SoilDisturbance`.

5. There are some E2E failues in `SoilDisturbance: Submission tests`, I expect that you'll probably resolve those when you update that page.

braughtg:
@Shahir-47 I just built a new sample DB release that includes a field J that has no beds and no active plant assets.  That should be helpful for your testing.  You can use the command `installDB.bash` to install the latest Sample Database release.

braughtg:
@Shahir-47 I should have also said that you will need to update your FarmData2 `development` branch.  I patched the test suite so that everything passes with the new field included.

Shahir-47:
@braughtg,

I wanted to let you know that I’ve introduced a new `requiredRow` prop to the `ActivePlantAssetPicklist` component. It’s now wired into the `picklistRequired` computed property so that `picklistRequired = required || requiredRow`. I also updated the logic so that whenever `required` is false we always emit `valid(true)` on init and on any validity change.

On the testing side, I implemented the four tests we discussed for both styling and event emissions:

1. **required=false**: both BedPicker and PicklistBase emit valid(true) immediately
2. **required=true + pick a row**: emits valid(true) once a row is checked
3. **requiredRow=false + pick a non-picklist bed**: still emits valid(true)
4. **requiredRow=true + no selections**: emits valid(false)

braughtg:
@Shahir-47 The functionality and testing for the `ActivePlantAssetPicklist` looks great.  There are a few small things that could be touched up.  I am happy to make these into a separate issue if you'd like to just get onto integrating the changes into the Soil Disturbance page. Let me know if you want to continue on this issue or if I should move it to the board for the summer interns.

To see the touch ups, on the Examples page for the `ActivePlantAssetPicklist` component:
- If there are no beds in a location it would be nice to fully hide the `BedPicker` and just show the table.  Select location "A" to see this.  There is a blank box just above the table that seems to be from the `BedPicker`. Can we hide that blank box if there are no beds?
- If there are no active plant assets it would be nice to fully hide the table and just show the `BedPicker`.  Select location "H" to see this.  There are beds in H, but there are no active plant assets, so the table is empty.  Can we hide the table if there are no active plant assets?
- If there are no bed and no active plant assets, it would be nice to hide both the bed picker and the table.  Select location "J" to see this.  There are no beds or active plant assets in "J", so the blank box appears above the table and the empty table appears.  Can we hide both the blank box and the empty table if there are no beds and no active plant assets?

Shahir-47:
Thanks for the detailed feedback, @braughtg. All of those suggestions make sense. I won’t have time to work on this right now, so I think it’s better to move it to the summer interns. Sorry about that, and really appreciate all the guidance so far.

braughtg:
> Thanks for the detailed feedback, @braughtg. All of those suggestions make sense. I won’t have time to work on this right now, so I think it’s better to move it to the summer interns. Sorry about that, and really appreciate all the guidance so far.

@Shahir-47 No worries.  Thanks for all you've done with FarmData2 to date. Your contributions have been outstanding!  Feel free to jump back in anytime you have time and interest.

braughtg:
This work has been incorporated into issue #488.

PR #465, opened 2025-06-05
fix(comp): highlights selected rows in PicklistBase

### LINKED ISSUE #445: PicklistBase: Highlight selected rows
Have the selected rows highlighted in some way.  This would include any row that is checked if `useUnits` is `false`, or has a value selected if `useUnits` is true.  

For example:

Image

See the documentation for the Bootstrap Vue Next Table component. The `_rowVariant` attribute on the row might be something that is helpful here.

fejesievo:
**Pull Request Description**

Enables row highlighting on selection

Closes #445 
---

**Licensing Certification**

Co-authored-by: Niloy Saha <143445417+niloy-saha-123@users.noreply.github.com>

### DIFF
--- a/components/PicklistBase/PicklistBase.behavior.comp.cy.js
+++ b/components/PicklistBase/PicklistBase.behavior.comp.cy.js
@@ -401,7 +401,7 @@ describe('Test the PicklistBase component behavior', () => {
       .should('contain', '✅ All');
   });
 
-  it('checks that row highlights on event change', () => {
+  it('checks that row highlights when row is selected', () => {
     cy.mount(PicklistBase, {
       props: {
         rows: [
@@ -415,14 +415,22 @@ describe('Test the PicklistBase component behavior', () => {
       },
     });
 
+    cy.get('[data-cy="picklist-row-0"]').should(
+      'not.have.class',
+      'selected-row'
+    );
     cy.get('[data-cy="picklist-checkbox-0"]').click();
     cy.get('[data-cy="picklist-row-0"]').should('have.class', 'selected-row');
 
+    cy.get('[data-cy="picklist-row-1"]').should(
+      'not.have.class',
+      'selected-row'
+    );
     cy.get('[data-cy="picklist-checkbox-1"]').click();
     cy.get('[data-cy="picklist-row-1"]').should('have.class', 'selected-row');
   });
 
-  it('clears row selection when clicking on the same row', () => {
+  it('highlighting is cleared when row is deselected', () => {
     cy.mount(PicklistBase, {
       props: {
         rows: [
--- a/components/PicklistBase/PicklistBase.behavior.comp.cy.js
+++ b/components/PicklistBase/PicklistBase.behavior.comp.cy.js
@@ -400,4 +400,49 @@ describe('Test the PicklistBase component behavior', () => {
       .find('span')
       .should('contain', '✅ All');
   });
+
+  it('checks that row highlights on event change', () => {
+    cy.mount(PicklistBase, {
+      props: {
+        rows: [
+          { name: 'Item 1', quantity: 5, location: 'GHANA' },
+          { name: 'Item 2', quantity: 3, location: 'KENYA' },
+        ],
+        columns: ['name', 'quantity'],
+        labels: { name: 'Name', quantity: 'Quantity', location: 'Location' },
+        picked: new Map(),
+        showInfoIcons: true,
+      },
+    });
+
+    cy.get('[data-cy="picklist-checkbox-0"]').click();
+    cy.get('[data-cy="picklist-row-0"]').should('have.class', 'selected-row');
+
+    cy.get('[data-cy="picklist-checkbox-1"]').click();
+    cy.get('[data-cy="picklist-row-1"]').should('have.class', 'selected-row');
+  });
+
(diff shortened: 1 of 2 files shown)


braughtg (changes requested):
I think that what you have here can be simplified quite a bit.  I would suggest not introducing the new `selectedRows` array and removing all of the related changes.

Then have your code use the existing `pickedRows` array.  The i-th entry in `pickedRows` array indicates whether the i-th row in the table is selected. If units are being used then `pickedRows[i]` is the number selected, and if it is anything greater than 0 the row should be highlighted. If units are not being used then `pickedRows[i]` will be either 0 or 1. So I think you can just have your `getRowClass` method look at `pickedRows` and determine if the row should be highlighted or not then return the appropriate `class` to bind to the row.

You should also undo the changes to the `handleUnitsButton` method because they are not related to this PR.

components/PicklistBase/PicklistBase.vue
```diff
@@ -92,8 +93,9 @@
                 v-bind:name="'picklist-checkbox-' + i"
                 v-bind:key="'checkbox' + i"
                 v-bind:disabled="showOverlay != null"
-                v-bind:checked="pickedRows[i]"
+                v-bind:checked="pickedRows[i] > 0"
```
braughtg: Let's revert this line to what it was before.

components/PicklistBase/PicklistBase.vue
```diff
@@ -92,8 +93,9 @@
...
                 v-on:change="(state) => handleCheckboxChange(i, state)"
+                v-on:click.stop
```
braughtg: Remove this line as it seems to have no effect.

components/PicklistBase/PicklistBase.behavior.comp.cy.js
```diff
@@ -400,4 +400,49 @@ describe('Test the PicklistBase component behavior', () => {
...
+          { name: 'Item 1', quantity: 5, location: 'GHANA' },
+          { name: 'Item 2', quantity: 3, location: 'KENYA' },
+        ],
+        columns: ['name', 'quantity'],
+        labels: { name: 'Name', quantity: 'Quantity', location: 'Location' },
+        picked: new Map(),
+        showInfoIcons: true,
+      },
+    });
+
```
braughtg, line 417: add check that class is not on the row before the click.

components/PicklistBase/PicklistBase.behavior.comp.cy.js
```diff
@@ -400,4 +400,49 @@ describe('Test the PicklistBase component behavior', () => {
       .find('span')
       .should('contain', '✅ All');
   });
+
+  it('checks that row highlights on event change', () => {
```
braughtg: checks that row highlights when row is selected

components/PicklistBase/PicklistBase.behavior.comp.cy.js
```diff
@@ -400,4 +400,49 @@ describe('Test the PicklistBase component behavior', () => {
...
+    });
+
+    cy.get('[data-cy="picklist-checkbox-0"]').click();
+    cy.get('[data-cy="picklist-row-0"]').should('have.class', 'selected-row');
+
+    cy.get('[data-cy="picklist-checkbox-1"]').click();
+    cy.get('[data-cy="picklist-row-1"]').should('have.class', 'selected-row');
+  });
+
+  it('clears row selection when clicking on the same row', () => {
```
braughtg: highlighting is cleared when row is deselected

components/PicklistBase/PicklistBase.behavior.comp.cy.js
```diff
@@ -400,4 +400,49 @@ describe('Test the PicklistBase component behavior', () => {
...
+        columns: ['name', 'quantity'],
+        labels: { name: 'Name', quantity: 'Quantity', location: 'Location' },
+        picked: new Map(),
+        showInfoIcons: true,
+      },
+    });
+
+    cy.get('[data-cy="picklist-checkbox-0"]').click();
+    cy.get('[data-cy="picklist-row-0"]').should('have.class', 'selected-row');
+
```
braughtg, line 424: check here too.

braughtg (changes requested):

components/PicklistBase/PicklistBase.vue
fejesievo: changed back to original line

components/PicklistBase/PicklistBase.vue
fejesievo: removed line

components/PicklistBase/PicklistBase.behavior.comp.cy.js
fejesievo: added check that class isn't on row first before click

components/PicklistBase/PicklistBase.behavior.comp.cy.js
fejesievo: changed it title to "checks that row highlights when row is selected"

components/PicklistBase/PicklistBase.behavior.comp.cy.js
fejesievo: changed it title to "highlighting is cleared when row is deselected"

PR #466, opened 2025-06-06
feat(comp): creates an example page for TraySizeSelector

### LINKED ISSUE #464: Add example page for TraySize selector component
The `TraySizeSelector` does not currently have an example page in the `farm_fd2_examples` module.  All of the components in the `components/` directory should have an example page.  So we should create one for the `TraySizeSelector`.

A good way to start would be to copy the example page for `SelectorBase` and then augment it with the additional elements necessary to interact with the `TraySizeSelector`.  Looking at the example page for `EquipmentSelector` would be a good place to gain insight into how to extend the example page for `TraySizeSelector`.

In addition, the following files in the `farm_fd2_examples/module` directory will need to be updated to add the new entrypoint for the `TraySizeSelector` example page.
- `farm_fd2_examples.libraries.yml`
- `farm_fd2_examples.links.menu.yml`
- `farm_fd2_examples.routing.yml`

The code in those files should provide enough of an example to figure out what needs to be added to get the new `TraySizeSelector` example page to appear.

niloy-saha-123:
**Pull Request Description**

Created an example page for TraySizeSelector and also created a test file to check if the page exists.

Closes #464 

---

**Licensing Certification**

### DIFF
--- a/modules/farm_fd2_examples/src/entrypoints/tray_size_selector/App.vue
+++ b/modules/farm_fd2_examples/src/entrypoints/tray_size_selector/App.vue
@@ -150,7 +150,4 @@ export default {
 tray-size-selector-example-hack {
   display: none;
 }
-.text-danger {
-  color: red;
-}
 </style>
--- a/modules/farm_fd2_examples/src/entrypoints/tray_size_selector/App.vue
+++ b/modules/farm_fd2_examples/src/entrypoints/tray_size_selector/App.vue
@@ -1,0 +1,183 @@
+<template>
+  <h3>TraySizeSelector Example</h3>
+  <p>
+    TraySizeSelector is a component that allows the user to pick a tray size
+    from a drop down list. It dynamically fetches its options.
+  </p>
+
+  <hr />
+  <TraySizeSelector
+    id="tray-size-selector"
+    data-cy="tray-size-selector"
+    label="Select Tray Size"
+    invalid-feedback-text="Tray size selection cannot be empty."
+    v-bind:required="required"
+    v-bind:showValidityStyling="validity.showStyling"
+    v-model:selected="form.selected"
+    v-on:valid="(valid) => (validity.selected = valid)"
+    v-on:ready="handleTraySizeSelectorReady"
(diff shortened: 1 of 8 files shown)


modules/farm_fd2_examples/src/entrypoints/tray_size_selector/App.vue
```diff
@@ -0,0 +1,183 @@
...
+  <table class="example-table">
+    <thead>
+      <tr>
+        <th>Event/Status</th>
+        <th>Payload/Value</th>
+      </tr>
+    </thead>
+    <tbody>
+      <tr>
+        <td>ready (from TraySizeSelector)</td>
```
braughtg: Remove the rows for the `ready` and `error` events as they are not events that we can affect using the props that we can control from this page.  So they are not particularly helpful for manual testing.  This will also keep the example page consistent with the others (e.g. `CropSelector`, etc).  You can also remove the additional code for handling those events as well.

braughtg (changes requested):
Overall this look really good.  Just one comment in the code about removing some things that are not really necessary.

braughtg:
Also, please use __Closes #464__ in your PR body.  That will cause the associated issue to be closed when the PR is merged.  Just using __#464__ will not cause the issue to be closed.

modules/farm_fd2_examples/src/entrypoints/tray_size_selector/App.vue
```diff
@@ -0,0 +1,156 @@
...
+</script>
+
+<style>
+@import url('@css/fd2-examples.css');
+@import url('@css/fd2-mobile.css');
+
+tray-size-selector-example-hack {
+  display: none;
+}
+.text-danger {
```
braughtg: Remove this unused style.

braughtg (changes requested):

modules/farm_fd2_examples/src/entrypoints/tray_size_selector/App.vue
niloy-saha-123: I have removed it.

PR #480, opened 2025-06-10
refactor(lib): splits farmosUtil library into smaller files

### LINKED ISSUE #207: refactor(fd2): break farmosUtil library up into smaller modules
The `library/farmosUtil/farmosUtil.js` file should be broken down into smaller units and imported back into the `farmosUtil.js` file. 

The test files in `library/farmosUtil` give a good strategy for breaking up the `farmosUtil.js` file.  Each test file tests a logical subset of the larger library. The library itself should be broken down in a parallel way.  The code tested by each test file should be extracted from the `farmosUtil.js` file into its own `.js` file that is named using the same scheme as the test files (e.g. the code for the functions for beds should be named `farmosUtil.beds.js`).  The `farmosUtil.js` file should then re-export all of the imported functions. 

Break out the code for everything in `farmosUtil.js` from `clearCachedUsers` down.  The code above `clearCachedUsers` should remain in `farmosUtil.js`.

The following snippet gives a simplified example of how the code would be structured. `a` and `b` would be the new smaller `.js` files that are then exported by `lib.js` which would be the `farmosUtil.js` file.  The `use.js` file is showing how any other file could use the functions in `farmosUtil.js`.

```
// a.js
export function a() {
    console.log("a");
}

// b.js
export function b() {
    console.log("b");
}

// lib.js
export * from './a.js';
export * from './b.js';

//use.js
import * as lib from './lib.js'

lib.a();
lib.b();
```

When the refactoring is complete, there will be one `js` file for each `unit.cy.js` file and the `farmosUtil.js` file should include just re-export the content of those files. If done correctly we should not need to make any changes outside of the `library/farmosUtil` directory.

niloy-saha-123:
Refactored the farmosUtil.js library into separate smaller more focused files and brings them back together in farmosUtil.js as a barrel file.

Closes #207 
---

**Licensing Certification**


braughtg (changes requested):
This PR is looking pretty good.  There are a few things that will need to be touched up:
- [ ] The new `makeDocks.bash` script requires each `.js` file other than `farmosUtil.js` to have a normal block comment at the top, with the first line of the comment being a one line description of the file. See the `farmosUtil.beds.js`, `farmosUtil.crops.js` and `farmosUtil.equipment.js` files for examples.  If a file really needs more than one line of comment, leave a blank after the one-line description and then add more.
- [ ] Rename the tests related to `farmosUtil.core.js` to be `farmosUtil.core.blahblah.unit.cy.js`. This will keep them near the file they are testing. Double check that this is the case for all of the other `.js` and corresponding `unit.cy.js`files as well.
- [ ] Restore the `farmosUtil.soilDisturbanceTerminationLog.unit.cy.js` file.  It contains only a small change unrelated to this PR.  So let's leave that change out, unless other changes are made to this file.
- [ ] Restore the large comment to the top of the `farmosUtil.js` file.  This comment had the description of the module and gave some examples of use.  You can copy this comment from the `farmosUtil.js` file in the current `development` branch.

PR #481, opened 2025-06-12
fix(fd2): removes unnecessary quantity deletes

### LINKED ISSUE #363: lib.js: Check for log deletion in `undo` before deleting quantities
Each of the entry points in the `farm_fd2` module (e.g. `tray_seeding`, `transplanting`,...) contains a `lib.js` file that manages the creation of the assets, logs and quantities needed to represent the operation.  Each log, asset or quantity is created in an operation for `farmosUtil.runTransaction` that has a `do` and `undo` operation.  The `do` attempts to create the asset, log, or quantity and the `undo` attempts to cleanup if the `do` or any subsequent `do` fails.

The caveat here is that when a log is deleted, farmOS automatically deletes the quantities that are associated with the log. There are 2 cases:

Case 1: A log referencing quantities is deleted:
- Some quantities are created
- A log referencing those quantities is created
- A later `do` operation fails
- `runTransaction` will then attempt to `undo` everything that was done.
  - the log will be deleted by its `undo`
  - the `undo` for each of the quantities will attempt to delete the quantity
    - These will fail because the deletion of the log already deleted the quantities.
    - __In this case it would be better not to do these deletes__
    - So the quantity should be deleted only if we know it exists.
  
Case 2: The creation of one of the quantities or the log that references the quantities fails
- Some quantities are created
- The creation of another quantity or the log referencing the quantities fails
- `runTransaction` will then attempt to `undo` everything that was done.
  - In this case, the quantities will need to be deleted by their `undo`

Case 1 is the one of concern here.  To handle this case each `undo` operation for a quantity needs to check to see if the log that referenced them was successfully deleted and if so then it should not attempt to delete the quantity.  Otherwise is should delete the quantity.

Currently some of the operations perform this check (e.g. see the 'bedFeetQuantity' in `lib.js` in `direct_seeding`) while others do not (e.g. see `depthQuantity` in `lib.js` in `soil_disturbance`).  

This issue asks that the:
- The `undo` for every quantity check if the log that references it was deleted (i.e. was `undone` prior to attempting to deleting the quantity.
- The associated `lib.*.unit.cy.js` tests be updated as necessary.
- Any e2e tests be updated as necessary.

'bedFeetQuantity' in `lib.js` in `direct_seeding` can be used as a model for others.

The `lib.js` in every module should be checked:
- [ ] Check quantity `undo` operations in `tray_seeding/lib.js`
- [ ] Check quantity `undo` operations in `direct_seeding/lib.js` 
- [ ] Check quantity `undo` operations in `transplanting/lib.js` 
- [ ] Check quantity `undo` operations in `cover_crop_seeding/lib.js` 
- [ ] Check quantity `undo` operations in `soil_disturbance/lib.js` 

Comment on this issue to request to work on one of the above tasks.  A new issue will be created for that task and it will be assigned to you.

fejesievo:
**Pull Request Description**

Added if conditionals to the lib file in cover crop seeding and tray seeding files, and added more units tests in each of their submitError test files. 

Closes #363 

---

**Licensing Certification**


modules/farm_fd2/src/entrypoints/tray_seeding/lib.submitError.unit.cy.js
```diff
@@ -85,4 +85,54 @@ describe('Test error when submitting tray seeding lib', () => {
       );
     }
   );
+
+  it(
```
braughtg: This test seems to be testing the same thing as the test above.  For tray seeding, because there is only one log created I think we can only have the test above.  The other test we would like to do is to have something after the seeding log creation fail.  That way when the seeding log is deleted we could check that `undo` for the quantities did not attempt to delete them.  But since the seeding log is the last thing created there is no way to do that here.

If you agree this test is the same as above, it can be removed.  Otherwise we should discuss it further.

modules/farm_fd2/src/entrypoints/cover_crop_seeding/lib.submitError.unit.cy.js
```diff
@@ -136,4 +136,153 @@ describe('Error when submitting using the cover_crop lib.', () => {
...
+
+  it(
+    'Verify 1 quantity is deleted when seedingLog is deleted with a (401 on the first post request)',
+    { retries: 4 },
+    () => {
+      let postRequestCount = 0;
+
+      cy.intercept('POST', '**/api/log/activity', (req) => {
+        postRequestCount += 1;
+        if (postRequestCount === 1) {
```
braughtg: No need for an if statement here.  Just failing the request will fail the first one which stops the transaction and calls all the `undo`'s.

braughtg (changes requested):
The tray seeding entry point looks pretty close.

When I look at `lib.js` in cover crop seeding, direct seeding, soil disturbance, and transplanting there are quite a few quantity `undo`s that still do not have the necessary if statements in them.

There are also some comments directly in the diff that you'll be able to see.

braughtg:
Be sure to uses "Closes #363" in your PR body instead of just "#363".  Without the "Closes" part GitHub will not automatically close the issue when the PR is merged.

Also, be sure to update the PR description once everything is completed.

modules/farm_fd2/src/entrypoints/tray_seeding/lib.submitError.unit.cy.js
fejesievo: Test is the same as the one above so I went ahead and removed it.

modules/farm_fd2/src/entrypoints/cover_crop_seeding/lib.submitError.unit.cy.js
fejesievo: Went ahead and removed the if statement since it isn't necessary.

bin/makeDocs.bash
braughtg, line 1: I think this got in here when we were doing some testing/debugging.  So you can checkout this file from the development branch which will remove it from the PR.

modules/farm_fd2/src/entrypoints/direct_seeding/lib.submitError.unit.cy.js
```diff
@@ -195,4 +195,55 @@ describe('Error when submitting using the direct_seeding lib.', () => {
         })
     );
   });
+
```
braughtg: I don't think we need this test because the case of attempting to delete those three logs is covered by the first test above.

modules/farm_fd2/src/entrypoints/soil_disturbance/lib.js
braughtg, line 1: There are missing checks in the Quantity `undo`s in this file.  I think we'll probably also need to add some tests to the `unit.cy.js` file.

modules/farm_fd2/src/entrypoints/transplanting/lib.submitError.unit.cy.js
```diff
@@ -192,4 +192,65 @@ describe('Error when submitting using the transplanting lib.', () => {
       );
     }
   );
+
```
braughtg: This looks like the same tests as above? Should it be being modified in some way to test something different?

modules/farm_fd2/src/entrypoints/tray_seeding/lib.js
braughtg, line 1: This looks good.  You'll just need to update the tests.

braughtg (changes requested):
This is getting there!  Keep at it!

modules/farm_fd2/src/entrypoints/soil_disturbance/lib.submitError.unit.cy.js
```diff
@@ -167,163 +172,165 @@ function runTest(activePlantAsset) {
...
-              req.reply({
-                statusCode: 401,
-              });
-            } else {
-              // Continue with the request normally for other requests
-              req.continue();
-            }
-          }).as('actPost');
+    it('Soil Disturbance: records are deleted if there is a submission error', () => {
+      //{ retries: 4 },
```
braughtg: Uncomment this line.

modules/farm_fd2/src/entrypoints/soil_disturbance/lib.submitError.unitCheck.unit.cy.js
```diff
@@ -0,0 +1,116 @@
+import { lib } from './lib';
+
+describe(`Test the Soil Disturbance lib submission `, () => {
```
braughtg: Can we give a more descriptive title here? Something still short but that gets at the difference between what this file is testing and what the other files are testing?

modules/farm_fd2/src/entrypoints/soil_disturbance/lib.submitError.unit.cy.js
```diff
@@ -167,163 +172,165 @@ function runTest(activePlantAsset) {
...
+            // Continue with the request normally for other requests
             req.continue();
-          }).as('quantDelete');
+          }
+        });
 
-          let activityLogDeletes = 0;
-          cy.intercept('DELETE', '**/api/log/activity/*', (req) => {
-            activityLogDeletes++;
-            req.continue();
```
braughtg, line 199: Should we still be failing these deletes so that we get the error messages here? That would be consistent with what is done in the else for the no active plant asset case.

braughtg (changes requested):
This looks really close... just a few small comments throughout the code for you to review.

We should remove `transplanting/lib.sample.unit.js`. It is leftover from the creation of the transplanting entry point and is not necessary.

modules/farm_fd2/src/entrypoints/soil_disturbance/lib.submitError.unit.cy.js
braughtg: If we fail these deletes, then there will be logs in the DB that will be referencing the plant asset that was created just for the test.  The test (in the `after`) then attempts to delete the plant asset that was created for the test, but this will fail because the logs created by the soil disturbance will refer to it.  So not failing these deletes ensure that the logs created by the soil disturbance will be deleted, and thus the `after` will be able to delete the plant asset created for the test.

PR #482, opened 2025-06-19
feat(fd2): Add multi-pass option to cover crop seeding form

### LINKED ISSUE #371: CoverCrop: add passes for application and incorporation equipment
The Cover Crop Seeding entrypoint includes `SoilDisturbance` components for Seed Application and for Seed Incorporation.  Currently these components do not include the "Passes" option. 

Image

Both of these should include the "Passes" option with a default value of 1 as is currently the case in the Soil Disturbance entrypoint.

Image

This will require:
- Study Cover Crop `lib.js` to understand the logs, assets and quantities that are created.
- Create a Cover Crop Seeding in farmOS and look at logs, assets, quantities in farmOS to see the relationship to `lib.js` 
- Study the Soil Disturbance entry point `lib.js` to understand the logs, assets, and quantities it creates, because it already handles passes.
- Create a Soil Disturbance in farmOS and look at logs, assets, quantities in farmOS to see the relationship to `lib.js`
- Changes the Cover Crop Seeding entry point (`App.vue`) to set the  `SoilDisturbance` component prop that includes passes in the UI for both seed application and seed incorporation.
  - See the documentation for the `SoilDisturbance` component for information about its props.
  - Also see the Soil Disturbance entrypoint for how it sets this prop.
- Update Cover Crop (`App.vue`) to add the value of passes to the `form` data.
  - Use Soil Disturbance as an example.
- Update Cover Crop e2e tests:
  - for `.submission` to check form
  - for `.seedApplicationEquipment` to ensure passes is showing.
  - for `.seedIncorporationEquipment` to ensure passes is showing.
- Change the Cover Crop Seeding `lib.js` to create an independent soil disturbance log for each pass.
  - See `lib.js` in the Soil Disturbance entrypoint to see how it handles multiple passes.
- Update unit tests for the Cover Crop Seeding `lib.js` to check that the changes for multiple passes is working correctly.
  - See the tests for `lib.js` in the Soil Disturbance entry point for examples of how it does its testing.
- Update the Cover Crop Seeding entry point e2e tests to include a test with multiple passes.

niloy-saha-123:
**Pull Request Description**

**Added multi-pass option to seeding form**

Closes #371 

---

**Licensing Certification**


modules/farm_fd2/src/entrypoints/cover_crop_seeding/lib.js
```diff
@@ -18,12 +18,12 @@ import * as farmosUtil from '@libs/farmosUtil/farmosUtil';
  *   seedApplicationSpeedQuantity: {quantity--standard},
  *   seedApplicationAreaQuantity: {quantity--standard},
  *   seedApplicationEquipment: [ {asset--equipment} ],
- *   seedApplicationActivityLog: {log--activity},
+ *   seedApplicationActivityLog: [ {log--activity} ],
```
braughtg: These should use an index for the pass in the name of the operation. E.g. `seedApplicationActivityLog(i)`.  So for pass 2 this would be `seedApplicationActivityLog2`

modules/farm_fd2/src/entrypoints/cover_crop_seeding/lib.js
```diff
@@ -18,12 +18,12 @@ import * as farmosUtil from '@libs/farmosUtil/farmosUtil';
...
  *   seedIncorporationDepthQuantity: {quantity--standard},
  *   seedIncorporationSpeedQuantity: {quantity--standard},
  *   seedIncorporationAreaQuantity: {quantity--standard},
  *   seedIncorporationEquipment: [ {asset--equipment} ],
- *   seedIncorporationActivityLog: {log--activity},
+ *   seedIncorporationActivityLog: [ {log--activity} ],
```
braughtg: Similar to above.

modules/farm_fd2/src/entrypoints/cover_crop_seeding/lib.js
```diff
@@ -167,30 +167,38 @@ async function submitForm(formData) {
...
-        undo: async (results) => {
-          await farmosUtil.deleteSoilDisturbanceActivityLog(
-            results['seedApplicationActivityLog'].id
-          );
-        },
-      };
-      ops.push(seedApplicationActivityLog);
+      // Loop for the number of passes specified in the form data.
+      for (let i = 0; i < formData.seedApplicationPasses; i++) {
+        const seedApplicationActivityLog = {
```
braughtg, line 175: Need to create unique quantities for each pass.  So move the quantity creations inside the loop and use the loop index `i` as part of the operation names.  E.g. `seedApplicationDepthQuantity1`

modules/farm_fd2/src/entrypoints/cover_crop_seeding/lib.js
```diff
@@ -167,30 +167,38 @@ async function submitForm(formData) {
...
+          // The name for each operation in the transaction must be unique.
+          name: `seedApplicationActivityLog-${i}`,
+          do: async (results) => {
+            // To make the log name itself unique in farmOS, we append the pass number.
+            const logName = `Seed Application Pass ${i + 1}`;
```
braughtg: Check SoilDisturbance to keep log name here consistent with what was done there.

modules/farm_fd2/src/entrypoints/cover_crop_seeding/lib.js
```diff
@@ -167,30 +167,38 @@ async function submitForm(formData) {
...
+              formData.beds,
+              ['tillage', 'seeding_cover_crop'],
+              results.plantAsset,
+              [
+                results.seedApplicationDepthQuantity,
+                results.seedApplicationSpeedQuantity,
+                results.seedApplicationAreaQuantity,
+              ],
+              seedApplicationEquipmentAssets,
+              logName // Pass the unique name to the creation utility.
```
braughtg: This line provides the "Note" in the log.  Check SoilDisturbance and make this consistent with what was done there.

modules/farm_fd2/src/entrypoints/cover_crop_seeding/lib.js
```diff
@@ -167,30 +167,38 @@ async function submitForm(formData) {
...
-            results['seedApplicationActivityLog'].id
-          );
-        },
-      };
-      ops.push(seedApplicationActivityLog);
+      // Loop for the number of passes specified in the form data.
+      for (let i = 0; i < formData.seedApplicationPasses; i++) {
+        const seedApplicationActivityLog = {
+          // The name for each operation in the transaction must be unique.
+          name: `seedApplicationActivityLog-${i}`,
```
braughtg: Check SoilDisturbance and use operation naming that is consistent with what was done there.

modules/farm_fd2/src/entrypoints/cover_crop_seeding/lib.js
```diff
@@ -252,44 +260,68 @@ async function submitForm(formData) {
...
-          );
-        },
-        undo: async (results) => {
-          await farmosUtil.deleteSoilDisturbanceActivityLog(
-            results['seedIncorporationActivityLog'].id
-          );
-        },
-      };
-      ops.push(seedIncorporationActivityLog);
+      // Loop for the number of passes specified in the form data.
```
braughtg: Similar comments to the SeedApplicationLog apply here.

modules/farm_fd2/src/entrypoints/cover_crop_seeding/lib.js
```diff
@@ -252,44 +260,68 @@ async function submitForm(formData) {
...
+              results[`seedIncorporationActivityLog-${i}`].id
+            );
+          },
+        };
+        ops.push(seedIncorporationActivityLog);
+      }
     }
 
     const result = await farmosUtil.runTransaction(ops);
+    // Initialize arrays to hold the multiple activity logs that are now created.
```
braughtg: With logs and quantities being named using the pass index these arrays will not be necessary.

modules/farm_fd2/src/entrypoints/cover_crop_seeding/lib.js
```diff
@@ -252,44 +260,68 @@ async function submitForm(formData) {
...
+
     if (seedApplicationEquipmentAssets.length > 0) {
       result['seedApplicationEquipment'] = seedApplicationEquipmentAssets;
     } else {
       result['seedApplicationEquipment'] = null;
       result['seedApplicationDepthQuantity'] = null;
       result['seedApplicationSpeedQuantity'] = null;
       result['seedApplicationAreaQuantity'] = null;
-      result['seedApplicationActivityLog'] = null;
+      //result['seedApplicationActivityLog'] = null;
```
braughtg: Need to understand why these are here and how they are used.  Then decide what to do with the quantities and logs which would have a passes index on the end. E.g. `seedApplicationDepthQuantity1` for the first pass.

Seems likely to have to do with testing?

modules/farm_fd2/src/entrypoints/cover_crop_seeding/lib.submit.unit.cy.js
braughtg, line 1: We should test the following using separate test files:
- Submission with no equipment and thus no passes (`lib.submit.noequipment.unit.cy.js)
  - Modify this file. 
- Submission with one pass (`lib.submit.onepass.unit.cy.js`)
  - Create new test file.
- Submission with multiple passes (`lib.submit.multiplepasses.unit.cy.js`)
  - Create new test file.

braughtg (changes requested):

modules/farm_fd2/src/entrypoints/cover_crop_seeding/lib.submit.multiplepasses.unit.cy.js
```diff
@@ -0,0 +1,89 @@
...
+    date: '1950-01-02',
+    crops: ['BEAN', 'CARROT'],
+    location: 'ALF',
+    beds: ['ALF-1', 'ALF-3'],
+    areaSeeded: 50,
+    seedApplicationEquipment: ['Tractor'],
+    seedIncorporationEquipment: ['Rake'],
+    seedApplicationDepth: 6,
+    seedApplicationSpeed: 5,
+    seedApplicationPasses: 3,
```
braughtg: Let's just do 2 passes unless there is a good reason to do 3.  This will make the test faster.

modules/farm_fd2/src/entrypoints/cover_crop_seeding/lib.submit.multiplepasses.unit.cy.js
```diff
@@ -0,0 +1,89 @@
...
+    seedIncorporationDepth: 8,
+    seedIncorporationSpeed: 3,
+    seedIncorporationPasses: 3,
```
braughtg: Two passes here also.

modules/farm_fd2/src/entrypoints/cover_crop_seeding/lib.submit.multiplepasses.unit.cy.js
```diff
@@ -0,0 +1,89 @@
...
+    winterKill: true,
+    winterKillDate: '1950-12-31',
+    comment: 'Multi-pass test',
+  };
+  let results = null;
+
+  before(() => {
+    const timeout = { timeout: 30000 };
+    cy.wrap(farmosUtil.getBedNameToAssetMap(), timeout)
+      .then(() => {})
```
braughtg: Most of the statements below load maps of information that were used in the original `submit` test.  But it doesn't look like those maps are being used in this test.  They can be removed.

modules/farm_fd2/src/entrypoints/cover_crop_seeding/lib.submit.unit.cy.js
```diff
@@ -12,62 +12,47 @@ describe('Submission using the cover_crop lib.', () => {
...
-
-  afterEach(() => {
-    cy.saveLocalStorage();
-    cy.saveSessionStorage();
+    const timeout = { timeout: 30000 };
+    cy.wrap(farmosUtil.getBedNameToAssetMap(), timeout)
+      .then((map) => {
+        bedMap = map;
+      })
+      .then(() => cy.wrap(farmosUtil.getLogCategoryToTermMap(), timeout))
```
braughtg, line 32: What is the reason that these were changed to be chained with `.then`?

modules/farm_fd2/src/entrypoints/cover_crop_seeding/lib.submit.unit.cy.js
braughtg, line 1: I see that you simplified the way the assertions were being done, which is nice.  But it also seems as if quite a number of assertions have been removed from this file.  Why were those removed?

modules/farm_fd2/src/entrypoints/cover_crop_seeding/lib.submitError.unit.cy.js
braughtg, line 1: Quite a number of assertions have been removed from this file.  It no longer seems to check that the correct number of delete events happen when there is a failure during the submission. Ideally this test will be more similar to its original form which tested more thoroughly.

modules/farm_fd2/src/entrypoints/cover_crop_seeding/lib.submitOptional.unit.cy.js
```diff
@@ -1,121 +1,142 @@
 import { lib } from './lib.js';
+import * as farmosUtil from '@libs/farmosUtil/farmosUtil';
 
-describe('Submission using the cover_crop lib.', () => {
-  beforeEach(() => {
-    cy.restoreLocalStorage();
-    cy.restoreSessionStorage();
-  });
+describe('Submission scenarios for cover crop lib', () => {
```
braughtg: A more descriptive title might be something like "Submissions with optional content omitted"

modules/farm_fd2/src/entrypoints/cover_crop_seeding/lib.submitOptional.unit.cy.js
```diff
@@ -1,121 +1,142 @@
...
+        results = res;
+      });
+    });
+
+    it('should create all standard logs', () => {
+      expect(results).to.have.property('plantAsset');
+      expect(results).to.have.property('seedingLog');
+      expect(results).to.have.property('seedApplicationActivityLog0');
+    });
+
```
braughtg, line 58: It would be a good idea to check for the quantities as well.  I think that all of the `expects' in this describe, including new ones for the quantities can be contained in a single `it`.

modules/farm_fd2/src/entrypoints/cover_crop_seeding/lib.submitOptional.unit.cy.js
```diff
@@ -1,121 +1,142 @@
...
+    before(() => {
+      cy.wrap(lib.submitForm(form), { timeout: 30000 }).then((res) => {
+        results = res;
+      });
+    });
+
+    it('should NOT create seed application records', () => {
+      expect(results).to.not.have.property('seedApplicationActivityLog0');
+      expect(results).to.not.have.property('seedApplicationDepthQuantity0');
```
braughtg: Why not check all of the quantities here?

modules/farm_fd2/src/entrypoints/cover_crop_seeding/lib.submitOptional.unit.cy.js
```diff
@@ -1,121 +1,142 @@
...
+    });
+
+    it('should still create seed incorporation records with correct details', () => {
+      const log = results.seedIncorporationActivityLog0;
+      expect(log.type).to.equal('log--activity');
+      expect(log.relationships.equipment[0].id).to.equal(
```
braughtg, line 134: This should also check for the quantities.

modules/farm_fd2/src/entrypoints/cover_crop_seeding/lib.submitOptional.unit.cy.js
```diff
@@ -1,121 +1,142 @@
...
-      expect(res.seedApplicationDepthQuantity).to.exist;
-      expect(res.seedApplicationSpeedQuantity).to.exist;
-      expect(res.seedApplicationAreaQuantity).to.exist;
-      expect(res.seedApplicationActivityLog).to.exist;
-      expect(res.seedIncorporationDepthQuantity).to.be.null;
-      expect(res.seedIncorporationSpeedQuantity).to.be.null;
-      expect(res.seedIncorporationAreaQuantity).to.be.null;
-      expect(res.seedIncorporationActivityLog).to.be.null;
+  // No equipment at all
+  describe('Submission without equipment', () => {
```
braughtg: Seems like this test should be checking for no seed incorporation equipment instead of no equipment at all.

braughtg (changes requested):
I added a fair number of comments throughout this PR.  Most are related to what seem to be removal of some things that had been being tested previously and what seemed like some unnecessary code.

I am also seeing two failing tests when I run:

`test.bash --e2e --fd2 --live --glob=modules/farm_fd2/**/cover_crop_seeding/*.e2e.cy.js`

modules/farm_fd2/src/entrypoints/cover_crop_seeding/lib.submit.multiplepasses.unit.cy.js
niloy-saha-123: Edited

braughtg:
The following tests also need to be updated:
- `cover_crop_seeding.submission.e2e.cy.js` - add checks for the values of the passes for both the application and the incorporation equipment.
- `cover_corp_seeding.submitReset.e2e.cy.js` - add check that reset also resets the values of the equipment selector (see `direct_seeding.submitReset.e2e.cy.js` for where it checks the soil disturbance component content after reset).

modules/farm_fd2/src/entrypoints/cover_crop_seeding/cover_crop_seeding.submitReset.e2e.cy.js
braughtg, line 1: In this file you'll need to make a few additional changes to fully test the form with the passes included.
- The `populateForm` function that enters data into the form in the browser should enter value in the passes field for both seed application and seed incorporation.  Use the way values are entered for speed and depth as examples.
- Add `it`s that check that invalid values for passes disables the submit button
  - for seed application
  - for seed incorporation
  - Use two separate `it`s for this.  Use the `its` that check that invalid depth values disable the submit button as examples and add the new `it`s just below the `it`s that check the corresponding depth values.
- It does not seem like the `it` for "Reset button resets form" should be passing.
  - You open the accordion but never enter any equipment.  If there is no equipment entered then the speed, depth and passes fields do not appear.  But you seem to be checking that they are visible.  They should not be until equipment is entered.  So something is off here if this test passes.  Check the `submitReset.e2e.cy.js` test in `direct_seeding` and note that it selects the `Tractor` after checking that the equipment was initially `null` (lines 250-262).
  - This `it` should also be testing that the depth, speed, passes fields are not visible when it checks that the equipment is `null` after the reset.  See lines 245-248 in `submitReset.e2e.cy.js` test in `direct_seeding`

braughtg (changes requested):
There are just a few more things that need to be tested in the `submitReset.e2e.cy.js` file. I added a comment there and we can also go over it when you come in this morning.

braughtg (changes requested):
Nice updates.  This is really close!

In `submission.e2e.cy.js`
  - enter distinct non-default values for passes in `submitForm` 
  - update the checks for the passes values to match in the first `it`
  - check for that the passes values are not reset during a sticky reset in the first `it`

In `lib.js`
  - The condition on the `if` in the `undo`s must include the pass index.
 
In `lib.submitError.unit.cy.js` 
  - change the number of passes to be 2 before submitting in the second and third `it`s  
  - update messages in `expect(...).to.contain` statements to use correct index.
  - Update failure counts in conditions on activity log intercepts 
    - 3 for seed application
   - 5 for seed incorporation

PR #487, opened 2025-06-21
Add @example JSDoc tag to runTransaction function

### LINKED ISSUE #221: Add @example code to `runTransaction` in `farmosUtil.js`
Add an `@example` JSDOC tag to the `runTransaction` function in `farmosUtil.js` that illustrates its use.  The example might just create a quantity and a log as operations in the transaction.

pranay-sankar-muppanenei:
Closes #221 

Enhanced the documentation for the [runTransaction](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-sandbox/workbench/workbench.html) function by adding an @example tag to demonstrate its usage with sample operations.

---

**Licensing Certification**

### DIFF
--- a/library/farmosUtil/farmosUtil.transaction.js
+++ b/library/farmosUtil/farmosUtil.transaction.js
@@ -28,68 +28,64 @@
  * If an operation was never attempted (and thus also not undone) the attribute for that operation will be `undefined`.
  *
  * @category Utilities
- * 
- * ```
- * {
- * name: string,
- * do: (Object) => async function performs an action and returns a result (e.g. creates a log, asset or quantity.) The argument has the same format as the return value. It contains the result of all prior operations, allowing them to be used by future operations.
- * undo: (Object) => async function that undoes the action performed by `do` (e.g. deletes a log, asset or quantity). The argument has the same format as the return value and contains the result of all successfully completed operations.
- * }
- * ```
- *
+ *  *
  * @example
  * // Create a transaction with a plant asset, a quantity, and a log
- * const operations = [
- *   {
- *     name: 'createPlantAsset',
- *     do: async () => {
- *       return await farmosUtil.createPlantAsset(
- *         '2023-10-01', // Example date
- *         'ZUCCHINI',   // Example crop name
- *         'Planting zucchini in the greenhouse' // Example comment
- *         [] // Example parents (empty array)
- *       );
- *     },
- *     undo: async (results) => {
- *       await farmosUtil.deletePlantAsset(results['createPlantAsset'].id);
- *     },
+ * let operations = [];
+ *
+ * const createPlantAsset = {
+ *   name: 'plantAsset',
+ *   do: async () => {
+ *     return await farmosUtil.createPlantAsset(
+ *       '2023-10-01', // Example date
+ *       'ZUCCHINI',   // Example crop name
+ *       'Planting zucchini in the greenhouse' // Example comment
+ *     );
  *   },
- *   {
- *     name: 'createQuantity',
- *     do: async () => {
- *       return await farmosUtil.createStandardQuantity(
- *         'count',
- *          50, // Example quantity value
- *         'Harvested Zucchini', // Example description
- *         'KG' // Example unit
- *          results.createPlantAsset, // Related asset
- *         'increment' // Inventory adjustment type  
- *       );
- *     },
- *     undo: async (results) => {
- *     if (results['createLog'] !== 'undone') { 
- *        await farmosUtil.deleteStandardQuantity(results['createQuantity'].id);
- *     }     
- *     },
+ *   undo: async (results) => {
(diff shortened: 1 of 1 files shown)


braughtg (changes requested):
@pranay-sankar-muppanenei Thanks for picking up this issue.  

You've got the structure right in this example, well done.  

I think I'd like to have an example that is a little more reflective of the use of the transactions in practice in the codebase.  I would suggest looking at `modules/farm_fd2_examples/src/entrypoints/example_entry_point/lib.js` for a simple example that creates a plant asset. And at `modules/farm_fd2/src/entrypoints/tray_seeding/lib.js` for an example of actual use that creates some quantitities and logs.  

I like the idea of keeping this simple, as you did. What do you think about maybe just creating one plant asset, one quantity and one log based on the above examples?

pranay-sankar-muppanenei:
@braughtg Thank you for the feedback! I’ve updated the `@example` tag for the 'runTransaction' function to reflect your suggestion. The example now creates one plant asset, one quantity, and one log, based on the patterns from 'example_entry_point/lib.js' and 'tray_seeding/lib.js'. It keeps the example simple while demonstrating practical usage. Let me know if there are any further adjustments you'd like!

library/farmosUtil/farmosUtil.transaction.js
```diff
@@ -28,6 +28,71 @@
...
+ *       );
+ *     },
+ *     undo: async (results) => {
+ *       await farmosUtil.deletePlantAsset(results['createPlantAsset'].id);
+ *     },
+ *   },
+ *   {
+ *     name: 'createQuantity',
+ *     do: async () => {
+ *       return await farmosUtil.createStandardQuantity(
```
braughtg: This does not look like a valid call to the `createStandardQuantity` function.  You can find that function in `library/farmosUtil/farmosUtil.quantity.js` and you can find its described in the documentation.  The documentation is generated as part of the install process for the development environment as described in `INSTALL.md`. You can also see a valid call for an example in `modules/farm_fd2/src/entrypoints/tray_seeding/lib.js` where it creates the `traysQuantity`starting on line 38.

library/farmosUtil/farmosUtil.transaction.js
```diff
@@ -28,6 +28,71 @@
...
+ *     undo: async (results) => {
+ *       await farmosUtil.deleteStandardQuantity(results['createQuantity'].id);
+ *     },
+ *   },
+ *   {
+ *     name: 'createLog',
+ *     do: async (results) => {
+ *       return await farmosUtil.createSeedingLog(
+ *         '2023-10-01', // Example date
+ *         'Greenhouse A', // Example location
```
braughtg: This should use one of the valid Greenhouse names that are populated in the Sample Database.  You can find a list of the greenhouses by visiting the Tray Seeding page in FarmData2 in the development environment and opening the Location drop down.

library/farmosUtil/farmosUtil.transaction.js
```diff
@@ -28,6 +28,71 @@
...
+ *     do: async () => {
+ *       return await farmosUtil.createStandardQuantity(
+ *         'count',
+ *         50, // Example quantity value
+ *         'Harvested Zucchini', // Example description
+ *         'KG' // Example unit
+ *       );
+ *     },
+ *     undo: async (results) => {
+ *       await farmosUtil.deleteStandardQuantity(results['createQuantity'].id);
```
braughtg: We have been working on a change in the way we `undo` quantities -  so the following would not have been in any of examples you've looked at yet. But we should include it here.

The delete of the quantity should be within with an `if` that checks if it needs to be deleted before attempting to delete it.  This check is helpful because when a log that references a quantity is deleted, the quantity is also deleted. So, if for example the seeding log created below were deleted by it's `undo`, then this quantity would have already been deleted.  Thus, attempting to delete it in its `undo` would fail.  So we can check first with:
```
if (results['createLog'] != 'undone') {
  // put the delete operation here.
}
```

library/farmosUtil/farmosUtil.transaction.js
```diff
@@ -28,6 +28,71 @@
  * If an operation was never attempted (and thus also not undone) the attribute for that operation will be `undefined`.
  *
  * @category Utilities
+ *
```
braughtg: It looks like there are lines here that duplicate what is above.  Be sure there is just one copy of them in the resulting file.  Keep the ones above and remove the ones below.

braughtg (changes requested):
Thanks for your continued effort on this issue. There are just a few more adjustments to be made as described in my other comments.

pranay-sankar-muppanenei:
Hi @braughtg, thanks for the suggestions! I’ve addressed all 4 points in the latest commit — including updating the greenhouse name to one from the Sample Database. Let me know if anything else needs adjustment.

PR #504, opened 2025-06-30
fix(comp): ActivePlantAssetPicklist - patch validity computation

### LINKED ISSUE #489: fix(comp): ActivePlantAssetPicklist validity
See #488 for instructions for fetching the existing feature branch before working on this issue.

The validity of the `ActivePlantAssetPicklist` is not being computed properly.

The `ActivePlantAssetPicklist` has two props `required` and `requiredRow` that will affect the validity computation for the component.  The validity should be based on these props and the currently selected beds and active plant assets.
- When `required` is true, the component should be valid if either a bed and/or an active plant asset in the picklist is selected. If `required` is true and nothing is selected then the component should not be valid.
- When `requiredRow` is true, the component should be valid only if an active plant asset in the picklist is selected.  If `requiredRow` is true and no active plant asset in the picklist is selected then the component should not be valid, even if an empty bed is selected.

The tests in `ActivePlantAssetPicklist.events.comp.cy.js` and `ActivePlantAssetPicklist.styling.comp.cy.js` will need to be updated accordingly.

The tests in `soil_disturbance.submitReset.e2e.cy.js` will need to be updated to ensure that the submit button is disabled when the `ActivePlantAssetPicklist` component is not valid.

In addition, the `ActivePlantAssetPicklist` Example entrypoint should be updated so that when `requiredRow` is true, `required` is forced to be true.

fejesievo:
**Pull Request Description**

Pull request for ActivePlantAssetPicklist Validity

Closes #489 

---

**Licensing Certification**

### DIFF
--- a/components/ActivePlantAssetPicklist/ActivePlantAssetPicklist.events.comp.cy.js
+++ b/components/ActivePlantAssetPicklist/ActivePlantAssetPicklist.events.comp.cy.js
@@ -102,7 +102,53 @@ describe('Test the ActivePlantAssetPicklist component events', () => {
     });
   });
 
-  it('3) with requiredRow=false and a non-picklist bed picked should emit valid=true', () => {
+
+  it('3) with required=true and a valid bed picked should emit valid=true', () => {
+    const readySpy = cy.spy().as('readySpy');
+    const validSpy = cy.spy().as('validSpy');
+
+    cy.mount(ActivePlantAssetPicklist, {
+      props: {
+        location: 'ALF',
+        required: true,
+        requiredRow: false,
+        onReady: readySpy,
+        onValid: validSpy,
+      },
+    }).then(() => {
+      // pick the first row
+
+      cy.get('[data-cy="picker-options"] input[value="ALF-2"]').click();
+      
+
+      cy.get('@validSpy').should('have.been.calledWith', true);
+    });
+  });
+
+
+  it('4) with required=true and an invalid bed picked should emit valid=false', () => {
+    const readySpy = cy.spy().as('readySpy');
+    const validSpy = cy.spy().as('validSpy');
+
+    cy.mount(ActivePlantAssetPicklist, {
+      props: {
+        location: 'ALF',
+        required: true,
+        requiredRow: false,
+        onReady: readySpy,
+        onValid: validSpy,
+      },
+    }).then(() => {
+      // pick the first row
+
+      cy.get('[data-cy="picker-options"] input[value="ALF-4"]').click();
+      
+
+      cy.get('@validSpy').should('have.been.calledWith', false);
+    });
+  });
+
+  it('5) with requiredRow=false and a non-picklist bed picked should emit valid=true', () => {
     const readySpy = cy.spy().as('readySpy');
     const validSpy = cy.spy().as('validSpy');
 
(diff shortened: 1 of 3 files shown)


components/ActivePlantAssetPicklist/ActivePlantAssetPicklist.vue
```diff
@@ -512,12 +535,13 @@ export default {
     },
 
     isValid() {
-      /**
-       * The validity of the bedPicker or PicklistBase has changed.
-       * @property {boolean} event whether the selections are valid or not.
-       */
       this.$emit('valid', this.isValid);
     },
+    /**
```
braughtg: This comment must remain above the `emit` statement for the documentation system to pick it up when generating the documentation.

components/ActivePlantAssetPicklist/ActivePlantAssetPicklist.vue
```diff
@@ -164,7 +164,30 @@ export default {
...
+      return this.checkedBeds && this.checkedBeds.length > 0;
+    },
+
     isValid() {
+      if (!this.picklistValid) {
+        return false;
+      }
+
+      if (this.requiredRow) {
+        if (!this.hasSelectedPlantAssets) {
```
braughtg: This `if` statement is redundant with the return below.  If `this.hasSelectedPlantAssets` is `false` you return `false`, but that would be the same as just returning `this.hasSelectedPlantAssets`.

components/ActivePlantAssetPicklist/ActivePlantAssetPicklist.events.comp.cy.js
braughtg, line 1: There is some inconsistent spacing in this file. Please use "Format Document" in VSCode and commit the changes. That should have run with the commit hook (unless you used `no-verify` which should only be used sparingly as necessary).

braughtg (changes requested):
This seems close... There are a few comments in the code, and also there was a failed test in `ActivePlantAssetPicklist.events.comp.cy.js` when I ran all tests.  You'll want to check on that.

braughtg:
I think we should revisit the "Valid-state tests" that are done so that they align with the specification in Issue #489. The tests 1-7 that are there were written for a different set of criteria so should be discarded and a new set of tests should be created that cover the new validity criteria. 

In writing the new tests use a structure similar to:
```
  it('A sample test structure', () => {
    const readySpy = cy.spy().as('readySpy');
    const validSpy = cy.spy().as('validSpy');

    cy.mount(ActivePlantAssetPicklist, {
      props: {
        location: 'ALF',
        required: true,
        requiredRow: false,
        onReady: readySpy,
        onValid: validSpy,
      },
    });

    cy.get('@readySpy')
      .should('have.been.calledOnce')
      .then(() => {
        cy.get('[data-cy="picker-options"] input[value="ALF-1"]').click();
        // or use cy.get('[data-cy="picklist-checkbox-0"]').click(); to pick a row.

        cy.get('@validSpy')
          .should('have.been.calledTwice')  // or use calledOnce here.
          .its('secondCall.args.0')  // and firstCall here.
          .should('equal', true);
      });
  });
```

If the result is to be `false` then there should be only one call to `validSpy`.  If the result is to be `true` then there should be two and the second one should be `true`.

braughtg:
There are also a few other items in #489 that need to be addressed before this is complete.

Issue #495, opened 2025-06-25
fix(fd2): Soil Disturbance - ensure submitted form includes correct beds


braughtg:
See #488 for instructions for fetching the existing feature branch before working on this issue.

Currently the value of `form.beds` is set only from the `LocationSelector` but earlier changes to the Soil Disturbance change in #492 mean that beds will be set from the `ActivePlantAssetPicklist`.

The page needs to be updated so that the value of `form.beds` is set correctly.

The tests in `soil_disturbance.submission.e2e.cy.js` will need to be checked that theytest that the correct beds are passed though to the `lib.submitForm` function.

braughtg:
It seems as if PR #505 will change the Soil Disturbance page to hide the `BedPicker` in the `LocationSelector` and use only the `BedPicker` in the `ActivePlantAssetPicklist`.  Thus, completing that PR will require a number of changes to the code and associated tests as well. Those changes will very likely complete this issue in the process.

PR #505, opened 2025-07-02
fix(fd2): SoilDisturbance - select all beds if no active plant assets

### LINKED ISSUE #499: fix(fd2): Soil Disturbance - select all beds if no active plant assets in location
When a location with beds but no active plant assets (e.g. "H") is selected, all beds should be selected by default when the location is selected.  This seems to make sense that the entire field might be being plowed or something such when there are no active plant assets in any of the beds.  This also makes the bed selections consistent with the default area % of 100 in this case.

niloy-saha-123:
**Pull Request Description**

Select all beds by default when a location with beds but no active plant assets is chosen.

Closes #499 

---

**Licensing Certification**

### DIFF
--- a/components/LocationSelector/LocationSelector.events.comp.cy.js
+++ b/components/LocationSelector/LocationSelector.events.comp.cy.js
@@ -411,4 +411,25 @@ describe('Test the LocationSelector component events', () => {
         .should('have.been.calledWith', 'Unable to fetch locations.');
     });
   });
+
+  it('emits update:beds with full bed list as third argument', () => {
+    const updateBedsSpy = cy.spy().as('updateBedsSpy');
+    cy.mount(LocationSelector, {
+      props: {
+        includeFields: true,
+        onUpdate: { beds: updateBedsSpy },
+      },
+    });
+
+    // Select a location with beds (e.g., "H")
+    cy.get('[data-cy="selector-input"]').select('H');
+    cy.get('@updateBedsSpy').should('have.been.called');
+
+    cy.get('@updateBedsSpy').then((spy) => {
+      const args = spy.lastCall.args;
+      expect(Array.isArray(args[2])).to.be.true; // third arg is array
+      expect(args[2].length).to.be.greaterThan(0); // should have beds
+      expect(args[2]).to.include('H-1'); // example bed
+    });
+  });
 });
--- a/components/LocationSelector/LocationSelector.vue
+++ b/components/LocationSelector/LocationSelector.vue
@@ -418,14 +418,8 @@ export default {
          * The selected beds have changed.
          * @property {Array<string>} checkedBeds an array containing the names of the selected beds.
          * @property {number} totalBeds the total number of beds in the selected location.
-         * @property {Array<string>} allBeds the full list of beds for the selected location.
          */
-        this.$emit(
-          'update:beds',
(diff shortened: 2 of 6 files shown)


braughtg (changes requested):
This implementation looks good.  

You should merge in the changes to the `soil-disturbance-update` branch, which contains some conflicts that will need to be resolved.  In doing so, be sure that all tests still pass.

Then you'll need to add some tests for this PR in:
- `LocationSelector.events.comp.cy.js` to test the new property of the `update:beds` event.
- `SoilDisturbance.location.e2e.cy.js` to test that all beds are selected by default if there are beds but no active plant assets (e.g. "H"), and that they are not if there are beds but no active plant assets (e.g. "ALF").

braughtg:
@niloy-saha-123 After having worked with this myself a bit, I am not sure we will be able to resolve the issue of the flashing of the `BedSelector` when choosing a first location with active plant assets - at least not with the approach we had been taking. This is due to the issue we identified of the `ActivePlantAssetPicklist` taking time to make the API call to determine if there are plant assets at the location.

I think that the solution here is to change course slightly.  So rather than trying to switch between the `BedSelector` in the `LocationSelector` and the one in the `ActivePlantAssetPicklist` I think we can just always use the `BedSelector` in the `ActivePlantAssetPicklist`. 

This would mean setting the `allowBedSelection` prop in the `LocationSelector` to `false` so that it never appears.  Then modifying the code to always get the selected bed information for `form.beds` from the `ActivePlantAssetPicklist`.

braughtg:
I believe that with the change in approach suggested above, completing this work will also require the completion of #495 as well.

niloy-saha-123:
> @niloy-saha-123 After having worked with this myself a bit, I am not sure we will be able to resolve the issue of the flashing of the `BedSelector` when choosing a first location with active plant assets - at least not with the approach we had been taking. This is due to the issue we identified of the `ActivePlantAssetPicklist` taking time to make the API call to determine if there are plant assets at the location.
> 
> I think that the solution here is to change course slightly. So rather than trying to switch between the `BedSelector` in the `LocationSelector` and the one in the `ActivePlantAssetPicklist` I think we can just always use the `BedSelector` in the `ActivePlantAssetPicklist`.
> 
> This would mean setting the `allowBedSelection` prop in the `LocationSelector` to `false` so that it never appears. Then modifying the code to always get the selected bed information for `form.beds` from the `ActivePlantAssetPicklist`.



Hi Professor @braughtg, I have been trying to figure out this process but I have realised that if we only allow for the bedpicker from the activeplantasset logic, locations like H with no activeplantasset will not show and bed selector. I might be wrong or have misunderstood the approach you are mentioning but this is what I am thinking about, and I am quite unsure what to do, and about the issue #495 , am I supposed to shift to that issue first and work on that and then come back to this?

braughtg:
> have realised that if we only allow for the bedpicker from the activeplantasset logic, locations like H with no activeplantasset will not show and bed selector.

Good point.  This sounds like we'll need to add a `prop` to the `ActivePlantAssetPicklist` that indicates that it should show empty beds.  This would include beds like those in "H" but also "ALF-3" and "ALF-4". 

I added issue #513 to address this.  Work should shift to that issue and then when it is complete return to this issue.

PR #509, opened 2025-07-08
Picklist style

### LINKED ISSUE #491: fix(comp): PicklistBase styling is incorrect when not required
Currently the `PicklistBase` component renders a green outline when the component is not required, empty, no rows are selected and the `showValidityStyling` prop is set to `true`.

Image

However, other components do not render any validity styling when `showValidityStyling` is set to `true` but they are empty and not required.  This issue asks that the `PicklistBase` component render with a normal border when it is not required and is empty (no rows selected), regardless of the `showValidityStyling` value.  

Image

The styling should remain the same as it is currently in other situations:
- Required and empty - renders a red outline and help message when `showValidityStyling` is `true`
- Required and not empty - renders a green outline when `showValidityStyling` is `true`
- Continues to render a normal border any time `showValidityStyling` is `false`

fejesievo:
**Pull Request Description**

Fixed visual styling bug in PicklistBase component that would occur when showValidityStyling prop was switched to true. Green border would show when it should've been the normal border color.

Closes #491 

---

**Licensing Certification**


fejesievo:
Accidentally pushed it to development instead of soil-disturbance-updates

PR #514, opened 2025-07-10
added prop for includeemptybeds

### LINKED ISSUE #513: fix(comp): ActivePlantAssetPicklist - add prop for "includeEmptyBeds"
Add a a `prop` to the `ActivePlantAssetPicklist` component named `includeEmptyBeds`.  This prop should default to `true`, but can be set to `false` by the entry point using the component. 

When `includeEmptyBeds` is `false`, the `BedPicker` in the `ActivePlantAssetPicklist` will show only beds that contain active plant assets.  For example, if the location is "ALF" then only "ALF-1" and "ALF-2" will appear in the `BedPicker`.  "ALF-3" and "ALF-4" will not appear in the `BedPicker` because they do not contain any active plant assets.

When `includeEmptyBeds` is `true`, the `BedPicker` in the `ActivePlantAssetPicklist` will show all of the beds in the location, whether or not they contain any active plant assets.  For example, if the location is "H" then beds "H-1" and "H-2" will appear in the `BedPicker`.  Similarly, if the location is "ALF" then all four beds "ALF-1" through "ALF-4" will appear in the `BedPicker`.

This change may or will require:
- Adding and/or modifying tests in `ActivePlantAssetPicklist.content.cy.js` to check that the `prop` works as expected when set to `false`.
- Adding and/or modifying tests in `ActivePlantAssetPicklist.behavior.cy.js` to check that changing the `prop` from `true` to `false` or vice versa works correctly and that changing locations includes the correct beds.
- Possibly other test adjustments if tests are broken by the change.

In addition, the example page for the `ActivePlantAssetPicklist` should be updated to include the new `prop`.

niloy-saha-123:
**Pull Request Description**

Added includeEmptyBeds prop to ActivePlantAssetPicklist to control bed filtering
When includeEmptyBeds is false, only show beds with active plant assets. When true, show all beds in the location. Uses PickerBase directly for cleaner implementation.

Closes #513 

---

**Licensing Certification**

### DIFF
--- a/components/ActivePlantAssetPicklist/ActivePlantAssetPicklist.behavior.comp.cy.js
+++ b/components/ActivePlantAssetPicklist/ActivePlantAssetPicklist.behavior.comp.cy.js
@@ -232,4 +232,191 @@ describe('Test the ActivePlantAssetPicklist component behavior', () => {
         });
     });
   });
+
+  it('Should filter beds when includeEmptyBeds changes from true to false for ALF', () => {
+    const readySpy = cy.spy().as('readySpy');
+
+    cy.mount(ActivePlantAssetPicklist, {
+      props: {
+        location: 'ALF',
+        isInGround: true,
+        isInTrays: true,
+        includeEmptyBeds: true,
+        onReady: readySpy,
+      },
+    }).then(({ wrapper }) => {
+      cy.get('@readySpy')
+        .should('have.been.calledOnce')
+        .then(() => {
+          // Initially shows all beds (ALF-1, ALF-2, ALF-3, ALF-4)
+          cy.get(
+            '[data-cy="picker-options"] input[name="picker-options"]'
+          ).should('have.length', 4);
+          cy.get(
+            '[data-cy="picker-options"] input[name="picker-options"][value="ALF-3"]'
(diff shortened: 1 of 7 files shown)


components/ActivePlantAssetPicklist/ActivePlantAssetPicklist.behavior.comp.cy.js
braughtg, line 1: I think one more test here would be good.  One where you switch from ALF to another location with beds (e.g. CHUAU) while `includeEmptyBeds` is false.

braughtg (commented):
This looks really good so far.   Just a few things:
- There are some failing tests when I `runAllTests.bash`. You can see them by running:
  - `test.bash --unit --fd2 --glob=modules/farm_fd2/src/entrypoints/soil_disturbance/*.unit.cy.js`
  - `test.bash --e2e --fd2 --live --glob=modules/farm_fd2/src/entrypoints/soil_disturbance/*.selections.e2e.cy.js`
- The example page for the `ActivePlantAssetPicklist` should have a switch added to the "Component Props" section for the `includeEmptyBeds` prop that allows it to be set to either `true` or `false`.

